PR #1, opened 2015-12-17
SLN fix, detonatepower, chrono delivery, makes mod load

Iamgoofball:
Ports the following pull requests from the old repo for RA2:
https://github.com/Phrohdoh/OpenRA/pull/63
https://github.com/Phrohdoh/OpenRA/pull/53

Sidenote about #53:
I implemented the sound effects and chrono teleport animation required to finish that pull request up.

Demo of #63

This also fixes the mod for the latest bleed version, along with fixing the crash on hitting escape.

### DIFF
--- a/OpenRA.Mods.RA2/Traits/ChronoResourceDelivery.cs
+++ b/OpenRA.Mods.RA2/Traits/ChronoResourceDelivery.cs
@@ -22,10 +22,13 @@ namespace OpenRA.Mods.RA2.Traits
 	{
 		[Desc("The number of ticks between each check to see if we can teleport to the refinery.")]
 		public readonly int CheckTeleportDelay = 10;
+
 		[Desc("The visual effect to play upon teleporting.")]
 		public readonly string TeleportEffect = null;
+
 		[Desc("The palette for the teleport visual effect.")]
 		public readonly string TeleportPalette = null;
+
 		[Desc("The sound for the teleport.")]
 		public readonly string TeleportSound = null;
 
@@ -97,13 +100,13 @@ namespace OpenRA.Mods.RA2.Traits
 			if (pos.CanEnterCell(dest.Value))
 			{
 				if (Info.TeleportSound != null)
-				{
+
 					Game.Sound.Play(Info.TeleportSound, self.OccupiesSpace.CenterPosition);
-				}
+
 				if (Info.TeleportEffect != null && Info.TeleportPalette != null)
-				{
+
 					world.AddFrameEndTask(w => w.Add(new Explosion(w, self.OccupiesSpace.CenterPosition, Info.TeleportEffect, Info.TeleportPalette)));
-				}
+
 				self.CancelActivity();
 				self.QueueActivity(new SimpleTeleport(dest.Value));
 				self.QueueActivity(nextActivity);
--- a/OpenRA.Mods.RA2/Traits/ChronoResourceDelivery.cs
+++ b/OpenRA.Mods.RA2/Traits/ChronoResourceDelivery.cs
@@ -1,0 +1,121 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2015 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation. For more information,
+ * see COPYING.
+ */
+#endregion
+
+using System;
+using OpenRA.Mods.Common.Activities;
+using OpenRA.Mods.Common.Effects;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Activities;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA2.Traits
+{
+	[Desc("When returning to a refinery to deliver resources, this actor will teleport if possible.")]
+	public class ChronoResourceDeliveryInfo : ITraitInfo, Requires<HarvesterInfo>
+	{
+		[Desc("The number of ticks between each check to see if we can teleport to the refinery.")]
+		public readonly int CheckTeleportDelay = 10;
+		[Desc("The visual effect to play upon teleporting.")]
+		public readonly string TeleportEffect = null;
+		[Desc("The palette for the teleport visual effect.")]
+		public readonly string TeleportPalette = null;
+		[Desc("The sound for the teleport.")]
+		public readonly string TeleportSound = null;
+
+		public virtual object Create(ActorInitializer init) { return new ChronoResourceDelivery(init.Self, this); }
+	}
+
+	public class ChronoResourceDelivery : INotifyHarvesterAction, ITick
+	{
+		public readonly ChronoResourceDeliveryInfo Info;
+
+		CPos? dest = null;
+		Activity nextActivity = null;
+		int ticksTillCheck = 0;
+
+		public ChronoResourceDelivery(Actor self, ChronoResourceDeliveryInfo info)
+		{
+			Info = info;
+		}
+
+		public void Tick(Actor self)
+		{
+			if (dest == null)
+				return;
+
+			if (ticksTillCheck <= 0)
+			{
+				ticksTillCheck = Info.CheckTeleportDelay;
+
+				TeleportIfPossible(self);
+			}
+			else
+				ticksTillCheck--;
+		}
+
+		public void MovingToResources(Actor self, CPos targetCell, Activity next)
+		{
+			Reset();
+		}
+
+		public void MovingToRefinery(Actor self, CPos targetCell, Activity next)
+		{
+			if (dest != null && dest.Value != targetCell)
+				ticksTillCheck = 0;
+
+			dest = targetCell;
+			nextActivity = next;
+		}
+
+		public void MovementCancelled(Actor self)
+		{
+			Reset();
+		}
+
+		public void Harvested(Actor self, ResourceType resource) { }
+		public void Docked() { }
+		public void Undocked() { }
+
+		void TeleportIfPossible(Actor self)
+		{
+			// We're already here; no need to interfere.
+			if (self.Location == dest.Value)
+			{
+				Reset();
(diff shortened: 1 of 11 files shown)


audio/voices.yaml
```diff
@@ -1,28 +1,26 @@
...
+		Select: sea, seb, sec, sed
+		Move: moa, mob, moc
+		Action: ata, atb, atc
+		Feedback: fea, feb, fec
+		Die: dia, dib, dic, did
 		Zapped: igenzapa
 		Burned: ienadia #gotta put something as temp else game crashes
 		Melted: igenmela, igenmelb, igenmelc
-		# [EngAllPowerDown] Sounds = ienspow
+		# [EngAllPowerDown] Sounds = ienapow
```
abcdefg30: Please remove these trailing whitespaces.

Iamgoofball: I don't know what that is.

abcdefg30: Sorry for being unclear. I mean the spaces at the end of the line.

Iamgoofball: Oh, yeah, I'll remove those.

Iamgoofball:
Sorry about cramming so much stuff into 1 PR, but this should allow all previous contributors to get right back to work on RA2, including dll changes.

scshunt:
I would reques tthat the monolith commit be split up so that individual change can be tracked separately.

Iamgoofball:
That's what we're working on right now.

audio/voices.yaml
```diff
@@ -1,28 +1,26 @@
...
-	DefaultVariant: .wav
-	Voices:
-		Select: ienasea, ienaseb, ienasec, ienased
-		Move: ienamoa, ienamob, ienamoc
-		Action: ienaata, ienaatb, ienaatc
-		Feedback: ienafea, ienafeb, ienafec
-		Die: ienadia, ienadib, ienadic, ienadid
+EngineerVoice:
+	DefaultVariant: .wav
+	Prefixes:
```
GraionDilach: Do not do this, please. The Allied Engineer has more voice files than it's Soviet counterpart!

I know having two clones of the unit (and preventing their availability based on factions) is bad but taht's how RA2 circumvented that issue.

Iamgoofball: You can't have two infantry units with different engine names using the same shp. I tried. It didn't work, we'd need to ship sovietengineer.shp and alliedengineer.shp alongside the mod, and that'd be distributing EA assets illegally.

chrisforbes: Sure you can. It just requires slightly different setup in your sequences
file.

GraionDilach: In this case it's probably more straightforward if `RenderSprites`->`Image` is used and not duplicating the sequences.

MustaphaTR: > I know having two clones of the unit (and preventing their availability based on factions) is bad but taht's how RA2 circumvented that issue.

Just no, :-1:  to cloning the Engineer (also the Dog). We should do a code wise fix for it.

What do game do if a sound file not found? Crash, not play anything or just ignore it and play another one? Because it looks like game will search for non-existing "ienssed" from current yaml code in this PR.

GraionDilach: Yes, the game will search for ienssed and will not find it and won't play anything - sound.log gets a File not found error line. It's not stupid technically when you're working with your own stuff, but when you're working with the WW assets you can't do anything with it.

I understand it's not an elegant solution but __it's the best you can do__.

abcdefg30: I don't think so and am with @MustaphaTR here. IMHO we should comment that out with a `# TODO` and add support for faction specific voices upstream. Your solution will mean adding a second unit just to remove it again once this is fixed upstream.

GraionDilach: Faction specific voices are supported technically. The problem comes up when these voicesets have different amount of files. In this case - IIRC, I'm at work - is that `ienased` exists in the mixes but `ienssed` does not.

This isn't a bug, but a feature IMO.

MustaphaTR: Wait wait wait! There are 3 more voice lines for Soviet Engineer, we can use one of them. http://www.ppmforums.com/viewtopic.php?highlight=engineer&t=33239

GraionDilach: RA2 assets __cannot__ be uploaded to the repository due to the game ain't released as freeware!

MustaphaTR: Not sure are they really the games assets. They were unused. They were never included ingame.

GraionDilach: Technically they are still the game's assets. (Or beta Red Alert 3's since that's where those got extracted from.)

MustaphaTR: Whatever forget it. We need to fix the limitation of having same amount of voices for each side, not find a workaround.

GraionDilach: That limitation is IMO a feature not a bug one needs to fix IMO.

scshunt: I can understand that you might not want to fix it, but why on earth would it be a feature?

GraionDilach: Missing files are hard to find in this system already. Sound.log failure is hard to notice already and fixing this would mean even less chances to debug/notice missing soundfiles.

scshunt: You're assuming that a fix would make that harder; it doesn't have to. Show a little creativity! ;)

abcdefg30: @GraionDilach I'm talking about something like
```
Voiced:
	VoiceSet: AlliedEngineer
	FactionVoiceSets:
		allies: AlliedEngineer
		soviet: SovietEngineer
```

GraionDilach: @scshunt @MustaphaTR wants such missing sounds remapped to what exists. That would do render such missing sounds harder to find out.

@abcdefg30 That probably would work but I find that redundant next to the faction prefixes.

scshunt: The faction fixes which don't work?

abcdefg30: > @abcdefg30 That probably would work but I find that redundant next to the faction prefixes.

I actually agree there, but I think it's the best solution.

MustaphaTR: > @scshunt @MustaphaTR wants such missing sounds remapped to what exists. That would do render such missing sounds harder to find out.

That's not what i want. I think we need a way to set counts seperately. I didn't mean game should play others if sound isn't found.

OpenRA.Mods.RA2/Traits/ChronoResourceDelivery.cs
```diff
@@ -0,0 +1,121 @@
...
+			}
+
+			var pos = self.Trait<IPositionable>();
+			var world = self.World;
+			if (pos.CanEnterCell(dest.Value))
+			{
+				if (Info.TeleportSound != null)
+				{
+					Game.Sound.Play(Info.TeleportSound, self.OccupiesSpace.CenterPosition);
+				}
```
abcdefg30: Please remove the braces around the single-line statements here and below (and add a new line after each of the `if`s).

OpenRA.Mods.RA2/Traits/ChronoResourceDelivery.cs
```diff
@@ -0,0 +1,121 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA2.Traits
+{
+	[Desc("When returning to a refinery to deliver resources, this actor will teleport if possible.")]
+	public class ChronoResourceDeliveryInfo : ITraitInfo, Requires<HarvesterInfo>
+	{
+		[Desc("The number of ticks between each check to see if we can teleport to the refinery.")]
+		public readonly int CheckTeleportDelay = 10;
+		[Desc("The visual effect to play upon teleporting.")]
```
abcdefg30: Also add new lines above the `Desc`s (except the first one), please.

OpenRA.Mods.RA2/Traits/SupportPowers/DetonateWeaponPower.cs
```diff
@@ -0,0 +1,135 @@
...
+			Action detonateWeapon = () => self.World.AddFrameEndTask(w =>
+			{
+				info.WeaponInfo.Impact(Target.FromCell(w, order.TargetLocation), self, Enumerable.Empty<int>());
+			});
+			self.World.AddFrameEndTask(w => w.Add(new DelayedAction(info.ActivationDelay, detonateWeapon)));
+
+			if (info.CameraActor != null)
+			{
+				var camera = self.World.CreateActor(false, info.CameraActor, new TypeDictionary
+					{
```
abcdefg30: This has one level of indentation too much.

OpenRA.Mods.RA2/Traits/SupportPowers/DetonateWeaponPower.cs
```diff
@@ -0,0 +1,135 @@
...
+			{
+				var wsb = self.Trait<WithSpriteBody>();
+				wsb.PlayCustomAnimation(self, info.ActivationSequence);
+			}
+
+			var targetPosition = self.World.Map.CenterOfCell(order.TargetLocation);
+
+			Action detonateWeapon = () => self.World.AddFrameEndTask(w =>
+			{
+				info.WeaponInfo.Impact(Target.FromCell(w, order.TargetLocation), self, Enumerable.Empty<int>());
```
abcdefg30: You can drop the braces here:
```
Action detonateWeapon = () => self.World.AddFrameEndTask(w =>
	info.WeaponInfo.Impact(Target.FromCell(w, order.TargetLocation), self, Enumerable.Empty<int>()));
```

scshunt:
I pulled the fixes out into #11 and #12 in order to get things working while we wait on the rest of this PR.

Mailaender:
You will need to rebase as those have already been merged. See https://help.github.com/articles/about-git-rebase/

Mailaender:
Superseded by https://github.com/OpenRA/ra2/pull/156.

Mailaender:
No activity here. The fixes have been included elsewhere.

PR #6, opened 2015-12-22
Remove AnnounceOnKill from Kirov Airship

MustaphaTR:
Moved Phrohdoh/OpenRA#81 to here.

#1 had to be splited anyway. This can be deleted from there.

### DIFF
--- a/audio/voices.yaml
+++ b/audio/voices.yaml
@@ -269,7 +269,6 @@ KirovVoice:
 		Attack: vkirata, vkiratb, vkiratc, vkiratd
 		Feedback: vkirfea, vkirfeb, vkirfec
 		Die: vkirdia, vkirdib, vkirdic, vkirdid
-		Kill: vkirdia, vkirdib, vkirdic, vkirdid
 
 DemolitionTruckVoice:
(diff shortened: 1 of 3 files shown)


abcdefg30:
This looks good, but please add `DeathSounds:` to `^Aircraft` in `defaults.yaml`.

MustaphaTR:
But not all aircrafts have a death sound? But most of them i think so i'll.

abcdefg30:
I think all aircraft units (`kirovs`, `harriers` (`black eagels`) and `night hawks`) had death sounds.

MustaphaTR:
Paradrop Plane, Osprey and Hornets don't have a death sound, but hit(crash) sounds.

abcdefg30:
Those would get a `-DeathSounds:` then.

abcdefg30:
We can also add `DeathSounds:` only to the `kirov` (and later to the other units which need that, when we add them). I'm actually ok with both.

MustaphaTR:
Updated. Added `DeathSounds:` to `^Aircraft`.

abcdefg30:
:+1:

PR #7, opened 2015-12-22
Fix Defences' wrong Inherits:

MustaphaTR:
Moved Phrohdoh/OpenRA#76 to here.

#1 had to be splited anyway. This can be deleted from there.

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -455,7 +455,7 @@ gawall:
 		Type: Concrete
 
 gapill:
-	Inherits: ^Building
+	Inherits: ^SupportBuilding
 	Buildable:
 		Queue: Support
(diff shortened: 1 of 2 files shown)


abcdefg30:
`gagap` probably wants an `-AutoTarget:` then.

MustaphaTR:
Updated.

MustaphaTR:
In fact AAs are also not Auto Targeted in orginal games but not sure should i change this here, it always looked like bug to me.

abcdefg30:
> In fact AAs are also not Auto Targeted in orginal games but not sure should i change this here, it always looked like bug to me.

That has to be a bug. For me they always auto targeted.

abcdefg30:
:+1:

PR #8, opened 2015-12-22
Fixes Engineers to use proper RA2 capturing

Iamgoofball:
They now capture buildings they are inserted into instead of sabotaging them.


abcdefg30:
:+1:

PR #11, opened 2015-12-23
Build fixes.

scshunt:
Fix various aspects of the build. I pulled them out of #1 because the irrelevant changes in that PR are preventing the critical fixes from being merged.

Added a stub C# file which can be removed once we commit a real one.


abcdefg30:
Looks good to me. :+1: 
Although I don't like including a placeholder...

PR #14, opened 2016-01-09
Added Kremlin Walls

MustaphaTR:
adsiz


abcdefg30:
One small thing: The wall defines `Armor=wood`, but the default of `^Wall` is `Armor: Type: Concrete`.
So please just add the correct armor and this should be ready to go.

MustaphaTR:
Updated.

abcdefg30:
:+1:

PR #15, opened 2016-01-09
Added Civilian (Tech) Hospital

MustaphaTR:
Works a bit different than orginal (Infantry does not really enters it), flag is missing like the Oil Derrick and i didn't set the Selection Bounds.

### DIFF
--- a/rules/tech-structures.yaml
+++ b/rules/tech-structures.yaml
@@ -24,3 +24,20 @@ caoild:
 		Sequence: idle-pump
 	GivesBuildableArea:
 
+cahosp:
+	Inherits: ^TechBuilding
+	Building:
+		Footprint: xxxx xxxx xxxx xxxx xxxx xxxx
+		Dimensions: 6, 4
+	Tooltip:
+		Name: Civilian Hospital
+	Health:
+		HP: 800
+	RevealsShroud:
+		Range: 6c0
+	Armor:
+		Type: Concrete
+	GivesBuildableArea:
+	RepairsUnits:
+		StartRepairingNotification:
+	RallyPoint:
--- a/rules/defaults.yaml
+++ b/rules/defaults.yaml
@@ -300,6 +300,9 @@
 		VoiceSet: GIVoice
 	QuantizeFacingsFromSequence:
 		Sequence: stand
+	Repairable:
+		RepairBuildings: cahosp
+		Voice: Move
 
 ^CivilianInfantry:
 	Inherits: ^Infantry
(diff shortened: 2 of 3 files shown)


rules/tech-structures.yaml
```diff
@@ -24,3 +24,23 @@ caoild:
 		Sequence: idle-pump
 	GivesBuildableArea:
 
+cahosp:
+	Inherits: ^TechBuilding
+	Building:
+		Footprint: xxxx xxxx xxxx xxxx xxxx xxxx
+		Dimensions: 6, 4
+	ExternalCapturable:
+	ExternalCapturableBar:
```
abcdefg30: This and the `ExternalCapturable` above look bogus. The engineers use `Captures`.
I removed them from `caoild` in https://github.com/OpenRA/ra2/commit/81023c50732589276ebb50346e8bd7acc781a5e8 just now.

rules/tech-structures.yaml
```diff
@@ -24,3 +24,23 @@ caoild:
...
+	EngineerRepairable:
```
abcdefg30: Isn't this unnecessary, as `^Building` already defines it?

MustaphaTR: I copy-pasted both this and above from Oil Derrick. OK, i'll remove.

MustaphaTR:
Updated.

Mailaender:
Seeing the discussion at https://github.com/OpenRA/OpenRA/pull/10363 escalate, I think you should add the RA2 logic right away. See https://github.com/OpenRA/OpenRA/pull/3520 for reference.

abcdefg30:
>  I think you should add the RA2 logic right away

This should already be the RA2 logic. Or do you mean adding logic for healing infantry really *inside* (kind of as passenger) and for the delay mentioned below?

From [cnc wikia](http://cnc.wikia.com/wiki/Tech_Hospital_(Red_Alert_2)):
> Originally, a damaged infantry unit would enter the hospital, and would come out a short time later fully healed. Commanders saw the usefulness of this and had Engineers trained to capture the building as a center to heal injured troops.

I think the way it is at the moment is the correct one to go (for the RA2 mod). (Means: No global heal upgrade.)

Mailaender:
I guess this changed in YR then:

> All infantry units receive portable medical kits and extra armor to heal themselves. However, Elite infantry units already know basic field survival and can heal themselves without any medical attention (like the heroes, such as Tanya, or elite infantry). However, a Tech Hospital or Tech Machine Shop was required to allow all units to completely heal or repair themselves.

:+1: here then.

GraionDilach:
Yes, this is the RA2 logic. The global heal upgrade (in along with the Machine Shop doing this same on vehicles) were part of YR.

Mailaender:
Works as promised. :white_check_mark:

PR #16, opened 2016-01-09
Remove "~!infantry.cuba" from Crazy Ivan

MustaphaTR:
Ivan is buildable by all Soviets in orginal game.


abcdefg30:
Looks like it.

PR #17, opened 2016-01-09
Fix Side Specific Units available for Everyone.

MustaphaTR:
Prerequisite sources was using old `Race:` tag instead of `Factions:`

### DIFF
--- a/rules/allied-infantry.yaml
+++ b/rules/allied-infantry.yaml
@@ -63,7 +63,7 @@ dog:
 	WithInfantryBody:
 	RenderSprites:
 		Image: adog
-		RaceImages:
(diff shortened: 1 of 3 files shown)


abcdefg30:
Could you also fix https://github.com/OpenRA/ra2/blob/4ae47e6788a9a935244718745f500e5c9b062409/rules/allied-infantry.yaml#L66 while you are at it, please?

MustaphaTR:
Updated. I think it should be in a different commit so i'm not gonna rebase.

abcdefg30:
:+1: thanks, I wonder why this wasn't caught by the upgrade rule @Phrohdoh ran.

PR #18, opened 2016-01-09
Fix other Side's Naval Unit Cameos are visible.

MustaphaTR:
All except aegis was missing of "~" on their `Prerequisite:`.


abcdefg30:
That's what you get when working with the `enabletech` cheat all the time. Thanks.

PR #20, opened 2016-01-10
Added more Civilian Props

MustaphaTR:
Pipes, V3 Ammunition, Tires, Practice Target and Derelict Tank. V3 Ammo has different explosion like in orginal.
miscs

### DIFF
--- a/rules/civilian-props.yaml
+++ b/rules/civilian-props.yaml
@@ -42,9 +42,62 @@ camisc04:
 		Footprint: x
 		Dimensions: 1, 1
 	Tooltip:
-		Name: Mail Box
+		Name: Mailbox
 	Health:
 		HP: 10
+	Armor:
+		Type: Concrete
+
+camisc05:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Pipes
+	Health:
+		HP: 10
+	Armor:
+		Type: Concrete
+
+camisc06:
+	Inherits: ^CivBuilding
+	Building:
+		Footprint: xx
+		Dimensions: 2, 1
+	Tooltip:
+		Name: V3 Ammunition
+	Health:
+		HP: 10
+	Armor:
+		Type: Concrete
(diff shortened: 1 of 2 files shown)


sequences/civilian-structures.yaml
```diff
@@ -277,6 +277,60 @@ camisc04:
 		ShadowStart: 5
 		Offset: 0, -15
 
+camisc04:
```
Mailaender: > Exception of type `System.ArgumentException`: MiniYaml.Merge, duplicate values found for the following keys: camisc04

MustaphaTR: lol, it looks like i copied too much thing.

MustaphaTR:
Can non squash right now. I'm not on computer, possibly can't open it today. Possibly can do it tomorrow.

Phrohdoh:
When you take ingame screenshots (for PR purposes) when introducing new actors (any changing artwork, really) please turn on the terrain geometry overlay.

MustaphaTR:
Squashed.

rules/civilian-props.yaml
```diff
@@ -42,12 +42,74 @@ camisc04:
...
 	Health:
 		HP: 10
 	Armor:
 		Type: Concrete
 
+camisc05:
+	Inherits: ^CivBuilding
+	Building:
+		Footprint: x
+		Dimensions: 1, 1
```
abcdefg30: This shouldn't be needed as `^CivBuilding` inherits `^Building` and this has [this](https://github.com/OpenRA/ra2/blob/9ef4a12c66258e95093bf3e651d6866382a9046d/rules/defaults.yaml#L77-L79) already defined.

(Same for `camsc11` and `camsc12`.)

MustaphaTR:
Updated.

Phrohdoh:
This all looks sensible to me, but I have not tested it ingame.

Mailaender:
Thanks!

PR #22, opened 2016-01-10
Added the other Flags

MustaphaTR:
flags


GraionDilach:
These would greatly benefit from a single ^Flag actor having all the traits excluding the tooltip imo.

MustaphaTR:
You i right but i can't do it now. Possibly tomorrow, sorry.

MustaphaTR:
Updated.

GraionDilach:
:+1:

PR #23, opened 2016-01-10
Add "TechLevel: Unrestricted" to "LobbyDefaults:"

MustaphaTR:
I know we don't have tech levels here yet but it was showning `none` on tech levels if not selected.

### DIFF
--- a/mod.yaml
+++ b/mod.yaml
@@ -199,6 +199,7 @@ LobbyDefaults:
 	Shroud: true
 	Fog: true
 	ShortGame: false
+	TechLevel: Unrestricted
 
 ChromeMetrics:
 	./mods/ra2/metrics.yaml


Mailaender:
Probably bit rot. Thanks!

Phrohdoh:
Does this even work if a TechLevel with that ID isn't defined?

MustaphaTR:
It shouldn't, but they are defined, possibly copy-pasted from RA1.

abcdefg30:
We don't have a techlevel called `none`, so "Does this even work if a TechLevel with that ID isn't defined?" is true (as it didn't crash before this PR).

Mailaender:
https://github.com/OpenRA/ra2/blob/master/rules/player.yaml#L66

PR #24, opened 2016-01-10
Make more Buildings Toggle Powerable

MustaphaTR:
All buildings shown image is now can be powered down. Previously only radars was.
power down

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -524,6 +524,8 @@ nasam:
 	AutoTarget:
 	Power:
 		Amount: -50
+	RequiresPower:
+	CanPowerDown:
 
 gtgcan:
 	Inherits: ^VoxelLighting
@@ -572,6 +574,8 @@ gtgcan:
 		Type: grandcannon
 	Power:
 		Amount: -100
+	RequiresPower:
+	CanPowerDown:
 
 gaorep:
 	Inherits: ^Building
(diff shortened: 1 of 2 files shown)


Mailaender:
The tesla coil electricity just stops instead of showing no bolts. =)

Phrohdoh:
Why are we allowing this? TS was the *only* game to allow powering down.

MustaphaTR:
Why not? RA1 and D2K mods of OpenRA do have Toggle Power. And we have it on UI.

Mailaender:
https://github.com/OpenRA/OpenRA/wiki/FAQ#this-is-not-true-to-the-original

Phrohdoh:
This is not upstream RA.
"Why not?" and "to be different' are not justified reasons for changing things.
Powerdown completely changes the balance/meta of a game.

I'd really like to see justification for changes rather than enabling everything the engine can do, RA has been perverted enough by that, don't do the same with RA2.

Mailaender:
Well, you yourself added @x-a-n-a-x`s UI which added the powerdown button https://github.com/Phrohdoh/OpenRA/commit/106982c498908a50063ee28fa4bcde1be2bbd0f2

Phrohdoh:
There has been discussion about that too which eventually converged on "fix the UI in the future when the artist has time".

I don't see a button existing as a good argument for changing this, but will take that on the nose.
We'll update the UI when time permits.

Mailaender:
Feel free to send a pull request removing the button and the traits for power down.

Issue #13, opened 2016-01-09
Can't open the RA2 Mod


MustaphaTR:
It crashes on bleed related to sounds. This may be realted to .wav loader. Not only files in audio.bag, also themes causes crash. Here is the log:
```
Red Alert 2 Mod at Version {DEV_VERSION}
on map ba240d57ac307937bab3223ce71e0e37b84e3a9e (Blank by Matthias Mailänder).
Operating System: Windows (Microsoft Windows NT 6.1.7601 Service Pack 1)
Runtime Version: .NET CLR 4.0.30319.34209
Exception of type `System.IO.InvalidDataException`: umenucl1.wav is not a valid sound file!
   konum: OpenRA.Sound.LoadSound(String filename) c:\Users\Merve-Elif\Downloads\OpenRA-bleed\OpenRA.Game\Sound\Sound.cs içinde: satır 72
   konum: OpenRA.Exts.GetOrAdd[K,V](Dictionary`2 d, K k, Func`2 createFn) c:\Users\Merve-Elif\Downloads\OpenRA-bleed\OpenRA.Game\Exts.cs içinde: satır 130
   konum: OpenRA.Primitives.Cache`2.get_Item(T key) c:\Users\Merve-Elif\Downloads\OpenRA-bleed\OpenRA.Game\Primitives\Cache.cs içinde: satır 35
   konum: OpenRA.Sound.PlayPredefined(Ruleset ruleset, Player p, Actor voicedActor, String type, String definition, String variant, Boolean relative, WPos pos, Single volumeModifier, Boolean attenuateVolume) c:\Users\Merve-Elif\Downloads\OpenRA-bleed\OpenRA.Game\Sound\Sound.cs içinde: satır 365
   konum: OpenRA.Sound.PlayNotification(Ruleset rules, Player player, String type, String notification, String variant) c:\Users\Merve-Elif\Downloads\OpenRA-bleed\OpenRA.Game\Sound\Sound.cs içinde: satır 380
   konum: OpenRA.Mods.Common.Widgets.ButtonWidget.HandleMouseInput(MouseInput mi)
   konum: OpenRA.Widgets.Widget.HandleMouseInputOuter(MouseInput mi) c:\Users\Merve-Elif\Downloads\OpenRA-bleed\OpenRA.Game\Widgets\Widget.cs içinde: satır 369
[...]

PR #28, opened 2016-01-11
Added support for installing this into the support folder

Mailaender:
See https://github.com/OpenRA/OpenRA/pull/10232. Had to fix several unrelated things to get this to run.

Fixes https://github.com/OpenRA/ra2/issues/13.
Fixes https://github.com/OpenRA/ra2/issues/31.

### DIFF
--- a/mod.yaml
+++ b/mod.yaml
@@ -233,6 +233,8 @@ Fonts:
 
 SupportsMapsFrom: ra2
 
+SoundFormats: Aud, Wav
+
 SpriteFormats: ShpTS, TmpTS, ShpTD
 
 SpriteSequenceFormat: TilesetSpecificSpriteSequence
--- a/mod.yaml
+++ b/mod.yaml
@@ -9,19 +9,19 @@ RequiresMods:
 
 Folders:
 	.
-	./mods/ra2
-	./mods/ra2/bits
-	./mods/ra2/bits/cameos
-	./mods/ra2/bits/vehicles
-	./mods/ra2/bits/structures
-	./mods/ra2/bits/temperate
-	./mods/ra2/bits/themes
-	./mods/ra2/bits/animations
-	./mods/ra2/uibits
+	ra2:
+	ra2:bits
+	ra2:bits/cameos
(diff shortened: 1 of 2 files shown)


Phrohdoh:
> The latest development version...

Doesn't fly at all. Instead you should use a hash that you know it works on.
That was part of the entire reason for supporting support mods (requires: git-123456).

Mailaender:
Maybe for releases, but not for development.

Phrohdoh:
Yes, for both.
This way you know what others have updated to (you know what is/isn't supported) and you know what not to add.

Mailaender:
Now also takes the revert from https://github.com/OpenRA/OpenRA/pull/10486 into consideration.

abcdefg30:
But needs a rebase.

Mailaender:
Rebased.

abcdefg30:
Sorry all other guys, we need to get this in, it's just a PITA to run the mod otherways at the moment. You need to use https://github.com/OpenRA/OpenRA/pull/10482 as OpenRA version for now, since "normal" OpenRA has a bug crashing this mod.

Mailaender:
I like the idea of developing against a set version of the OpenRA engine and CI test this with the engine from DEBs installed although that will be slightly boring and slow down progress a lot. My suggestion would be master/next/bleed (stable/playtest/bleed) as branches if we can keep up with the additional overhead of maintaining 3 branches.

Phrohdoh:
I highly suggest against more than a single branch as we've hardly been able to handle that so far.
A single `master` that is set against the latest upstream `release-` is the best way forward in my opinion.

abcdefg30:
I think at least to branches would be good. Only relying on the release will slow down progress a lot (e.g. we have to wait for a new release until a required upstream feature can be used) and it'll also be really not too convenient to update huge chunks of the mod at once (when a new release is tagged) instead of doing small but steady updates.

Mailaender:
We could transition to developing against a playtest soon and a release afterwards I guess.

PR #29, opened 2016-01-11
Setup the undeploy cursor for construction yards

Mailaender:
At the moment it re-uses the deploy ones which is irritating.

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -27,6 +27,7 @@ gacnst:
 		IntoActor: amcv
 		Offset: 1,1
 		Facing: 96
+		DeployCursor: undeploy
 	ProductionBar@Building:
 		ProductionType: Building
 	ProductionBar@Support:
--- a/rules/soviet-structures.yaml
+++ b/rules/soviet-structures.yaml
@@ -29,6 +29,7 @@ nacnst:
 		IntoActor: smcv
 		Offset: 1,1
 		Facing: 96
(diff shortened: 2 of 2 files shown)


abcdefg30:
Sorry, but the cursor looks a bit weird, I think...

Mailaender:
I checked as I thought the same thing. It is the same one used in TS.

Phrohdoh:
Is it what vanilla RA2 used or did they name it something else?

GraionDilach:
RA2 never used an undeploy cursor. The irritating thing was the RA2 behaviour.

abcdefg30:
Sorry, but I don't think this is a good idea with the current cursor, it just doesn't look fitting imho.

abcdefg30:
Fun fact: RA does the same (which doesn't mean that we have to, too).

Anyway, what do you think about [this](https://www.dropbox.com/sh/ga3zn88ptg3j8qb/AADtsWdaq7BdJlgX6L6fJDrja?dl=0) undeploy cursor made by @x-a-n-a-x?

GraionDilach:
I like it but I guess @Phrohdoh will :-1: it imminently.

abcdefg30:
> 20:48:41 (Phroh|work) +1
20:48:47 (Phroh|work) but please try it ingame first
20:48:49 (Phroh|work) before merging

abcdefg30:
Superseded by #34.

PR #9, opened 2015-12-23
Adds the Soviet Sentry Gun.

Iamgoofball:
Adds the Soviet Sentry Gun base defense.

Pretty much just a pill box for the soviets, as it was in original RA2.

PR #30, opened 2016-01-11
Fixed the TEMPERATE map import and clean up the maps

Mailaender:
Somehow the tileset was called `TEMPERAT` while all the original maps define `TEMPERATE` compare with http://modenc.renegadeprojects.com/Theatres as well. I fixed that and replaced all junk maps with proper ones where no actors are missing using the latest version of the map importer https://github.com/OpenRA/OpenRA/pull/10397 which includes several fixes such as proper map boundaries and global lighting.


Phrohdoh:
This also adds maps (which the title does not mention) without screenshots, and I'm willing to bet that ore creation isn't correct on them (it hasn't been in the past unfortunately).

GraionDilach:
The ore issues are caused by Westwood mappers using third ore instead of first one.

See http://modenc.renegadeprojects.com/Image#In_YR and https://github.com/Phrohdoh/oramod-ra2/blob/master/extra-files/rules.ini#L1523 regarding the problem. (The TIB02 entries in art.ini shares the image with the TIB entries where's this confusion comes. ModEnc uses the internal indexes in the article, which starts from #0, the additional offset differences are caused by WW skipping index #40 and #41 in rules.ini).

These will probably need a fork of the TS map importer branch with #127-#139 also being converted to Ore being the sole change.

abcdefg30:
Amazonas delta looks good to me:
openra-2016-01-13t200641z
openra-2016-01-13t200645z
Although this looks like a small tile issue to me:
openra-2016-01-13t200649z
Snow Valley, Heck Freezes Over and Powder Keg look good to me, too:
openra-2016-01-13t200854z
openra-2016-01-13t201558z
openra-2016-01-13t201740z
Defcon 6 has no ore.

Mailaender:
> The ore issues are caused by Westwood mappers using third ore instead of first one.

Thanks for the hint. I assumed that it could be an error in the terrain tiles where resource growth is allowed as the drills also don't seem to work, but didn't investigate it further.

Mailaender:
Rebased to also account for recently added sentry turret. https://github.com/OpenRA/ra2/pull/9

GraionDilach:
Come to think of it, that's also possible, Ore and Gems are only allowed on Clear, while they were also available on Rough atleast - notsure regarding DirtRoad.

abcdefg30:
Sorry, but needs a rebase.

Mailaender:
Rebased and added the missing terrain type definition to fix resource seeding:

image

Mailaender:
> Amazonas delta looks good to me:
> [...]
> Although this looks like a small tile issue to me:
> [...]

I removed it again.

abcdefg30:
Looks good to me, now. :+1: / :white_check_mark:

PR #32, opened 2016-01-14
Add SoundOnDamageTransition to civilian props

abcdefg30:
Adds `SoundOnDamageTransition` where I found it.
rules.ini:
```
DamageSound=BuildingMetalDamaged
DieSound=BuildingMetalDamaged
```
and sound.ini:
```
[BuildingMetalDamaged]
Sounds=bmetdama bmetdamb bmetdamc
```

For some reason (the explosion sounds?) some of them only had `DamageSound` defined, so I kept it that way. In addition I did some cleanup and removed some values that are already inherited (see https://github.com/OpenRA/ra2/pull/20#discussion_r49741770).

Follow-up of #20.

### DIFF
--- a/rules/civilian-props.yaml
+++ b/rules/civilian-props.yaml
@@ -1,9 +1,5 @@
 camisc01:
 	Inherits: ^CivBuilding
-	Selectable:
-	Building:
-		Footprint: x
-		Dimensions: 1, 1
 	Tooltip:
 		Name: Barrels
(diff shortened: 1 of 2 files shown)


Mailaender:
Can't you just inherit this from the defaults?

abcdefg30:
Most of the civilian builings don't want those sounds. But I added them to `^Flag`.

PR #33, opened 2016-01-14
Fix the turret offsets of the Sentry Gun and the Patriot Missile System

abcdefg30:
I used those wiki pages to confirm the offsets:
http://cnc.wikia.com/wiki/Sentry_Gun_(Red_Alert_2)
http://cnc.wikia.com/wiki/Patriot_missile_system_(Red_Alert)

They might not be perfect though.

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -508,7 +508,7 @@ nasam:
 	Turreted:
 		ROT: 10
 		InitialFacing: 224
-		Offset: 900,0,0
+		Offset: 550,0,0
 	AttackTurreted:
 		Voice: Attack
 	Armament:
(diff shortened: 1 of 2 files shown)


Phrohdoh:
Screenshot please.
If this looks right to you I'd say just go ahead and merge it. :+1:

abcdefg30:
Hm, I'm not sure anymore if this is the correct way to do it, will have another look tomorrow.

Mailaender:
Patriot missile looks way better. :white_check_mark:

Phrohdoh:
The Patriot Missile System still looks too high to me.
Compare to the link you gave.

screen shot 2016-01-16 at 6 31 50 pm

Iamgoofball:
I feel like the sentry gun is a little squished in, but eh

:+1:

abcdefg30:
> Hm, I'm not sure anymore if this is the correct way to do it, will have another look tomorrow.

Doesn't look like I can do it really better. Didn't change anything now.

> The Patriot Missile System still looks too high to me. :-1:

The problem here is that the turret is too low then when facing the opposite side.

PR #34, opened 2016-01-15
Add an undeploy cursor and make construction yards use it

abcdefg30:
See #29.


cursors.yaml
```diff
@@ -101,7 +101,7 @@ Cursors:
 			Length: 9
 		deploy-blocked:
 			Start: 119
-		undeploy:
+		undeploy-unused:
```
Phrohdoh: Why keep unused markup?

abcdefg30: Removed it.

abcdefg30:
Updated again.

cursors.yaml
abcdefg30: Readded it as requested, just as `undeploy-silver` now, as it's color is silver. I tend to agree to @Mailaender that this a) won't hurt and b) the artwork is still there and could be used.

Mailaender:
Nicely done. Thanks to @x-a-n-a-x.

PR #35, opened 2016-01-15
Rename the thirdparty directory to dependencies

abcdefg30:
And add a make file. Unfortunately I don't know what linux/mac users need there, so the file is windows only at the moment. :(

### DIFF
--- a/OpenRA.Mods.RA2/dependencies/readme.txt
+++ b/OpenRA.Mods.RA2/dependencies/readme.txt
@@ -1,0 +1,5 @@
+You should copy following files out of your main OpenRA install in here:
+- OpenRA.Game.exe
+- OpenRA.Mods.Common.dll
+- OpenRA.Mods.RA.dll
+- OpenRA.Mods.TS.dll
--- a/make.ps1
+++ b/make.ps1
@@ -1,0 +1,130 @@
+function FindMSBuild
+{
+	$msBuildVersions = @("4.0")
+	foreach ($msBuildVersion in $msBuildVersions)
+	{
+		$key = "HKLM:\SOFTWARE\Microsoft\MSBuild\ToolsVersions\{0}" -f $msBuildVersion
+		$property = Get-ItemProperty $key -ErrorAction SilentlyContinue
+		if ($property -eq $null -or $property.MSBuildToolsPath -eq $null)
+		{
+			continue
+		}
+		$path = Join-Path $property.MSBuildToolsPath -ChildPath "MSBuild.exe"
+		if (Test-Path $path)
+		{
+			return $path
+		}
+	}
+	return $null
+}
+
+if ($args.Length -eq 0)
+{
+	echo "Command list:"
+	echo ""
+	echo "  all             Builds the mod dll."
+	echo "  dependencies    Copies the mod's dependencies into the"
+	echo "                  'OpenRA.Mods.RA2/dependencies' folder."
+	echo ""
+	$command = (Read-Host "Enter command").Split(' ', 2)
+}
+else
+{
+	$command = $args
+}
+
+if ($command -eq "all")
+{
+	$msBuild = FindMSBuild
+	$msBuildArguments = "/t:Rebuild /nr:false"
+	if ($msBuild -eq $null)
+	{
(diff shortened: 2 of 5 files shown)


Phrohdoh:
If another Windows user can confirm this works and isn't more than we need :+1: and I can write up a unix makefile.

Mailaender:
My suggestion would be to avoid all C# code in this repository. This is a step backwards in getting this to run quickly and it adds redundancy as we will manage dependencies here and in OpenRA/OpenRA.

OpenRA.Mods.RA2/dependencies/readme.txt
```diff
@@ -0,0 +1,5 @@
+You should copy following files out of your main OpenRA install in here:
+- OpenRA.Game.exe
+- OpenRA.Mods.Common.dll
+- OpenRA.Mods.RA.dll
+- OpenRA.Mods.TS.dll
```
Mailaender: https://github.com/OpenRA/ra2/pull/28 allows you to avoid that step.

OpenRA.Mods.RA2/dependencies/readme.txt
```diff
@@ -0,0 +1,5 @@
+You should copy following files out of your main OpenRA install in here:
```
Mailaender: https://github.com/OpenRA/ra2/pull/28 allows you to avoid that step.

Phrohdoh: That simply isn't true.
These assemblies are referenced by the RA2 csproj.

Phrohdoh:
The point isn't to get RA2 running quickly.
It is the development of a RA2 mod which requires a custom assembly.

I've brought up countless times turning release and playtest assemblies into NuGet packages just for this reason, but nothing fruitful has come of that.

Mailaender:
Sounds like a plan. Go for it. Publishing stuff on NuGet isn't hard and you can automate it conveniently with @AppVeyor.

OpenRA.Mods.RA2/dependencies/readme.txt
Mailaender: I see. Guess have to do a botched job with symlinks then.

Phrohdoh: You should actually copy the assemblies from the required version into this directory as instructed.

abcdefg30:
Updated and added checks against missing files / invalid paths as requested.

make.ps1
```diff
@@ -0,0 +1,130 @@
...
+	$OpenRADir = Read-Host "Enter the path to your OpenRA install"
+	if (!(Test-Path $OpenRADir))
+	{
+		echo "Given directory does not exist!"
+	}
+	else
+	{
+		cd $OpenRADir
+
+		if (!(Test-Path mods/ra/OpenRA.Mods.RA.dll))
```
RockyTV: @abcdefg30 You typed the wrong filename there.

OpenRA.Mods.RA2/dependencies/readme.txt
Mailaender: If you want to go that way, you have to make sure that the assemblies from OpenRA/OpenRA become versioned.

make.ps1
abcdefg30: Derp, fixed.

Mailaender:
I assume this won't break https://github.com/OpenRA/ra2/pull/45, but a rebase would be nice just to make sure. Can't otherwise really test this as I am on Linux.

OpenRA.Mods.RA2/dependencies/readme.txt
Phrohdoh: See [my NuGet comment](https://github.com/OpenRA/ra2/pull/35#issuecomment-172032814).

abcdefg30:
Rebased. For some reason I totally forgot to do this, sorry.

abcdefg30:
Merging with a :+1: from @RockyTV.
(`Jan 15 18:52:41 <RockyTV> abcdefg30, all in all, your script works`)

Mailaender:
Added something to https://github.com/OpenRA/ra2/wiki/ which probably should be improved.

PR #36, opened 2016-01-15
Fix crash at Map Editor

MustaphaTR:
`TWO_LINE_TOOLTIP` was missing and causing crash when hovered on an actor in UI of editor.

Just copy pasted from `ra`. Not sure how does it look but everything else also looks same with `ra` mod.

### DIFF
--- a/chrome/tooltips.yaml
+++ b/chrome/tooltips.yaml
@@ -9,6 +9,23 @@ Background@SIMPLE_TOOLTIP:
 			Height: 23
 			Font: Bold
 
+Background@TWO_LINE_TOOLTIP:
+	Logic: ButtonTooltipLogic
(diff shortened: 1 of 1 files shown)


Phrohdoh:
Currently the RA2 UI is an exact copy of RA, so this ought to look fine.
Could you include a screenshot, please?

MustaphaTR:
adsiz

Mailaender:
> `TWO_LINE_TOOLTIP` was missing and causing crash when hovered on an actor in UI of editor.

Thanks! This was very annoying.

> Currently the RA2 UI is an exact copy of RA, so this ought to look fine.

Why did you duplicate it then instead of referencing the original?

PR #40, opened 2016-01-16
Added Goodie Crates

MustaphaTR:
I couldn't get water image work. I need some help for it. Goodie types are just for existance of crate, it should be adjusted for orginal RA2 later.

### DIFF
--- a/rules/misc.yaml
+++ b/rules/misc.yaml
@@ -16,6 +16,38 @@ waypoint:
 	WithSpriteBody:
 	QuantizeFacingsFromSequence:
 
+crate:
+	Inherits: ^Crate
+	Crate:
+		Lifetime: 240
+	GiveCashCrateAction@500:
+		Amount: 500
+		SelectionShares: 10
+		UseCashTick: yes
+	GiveCashCrateAction@1000:
+		Amount: 1000
+		SelectionShares: 20
+		UseCashTick: yes
+	GiveCashCrateAction@1500:
+		Amount: 1500
+		SelectionShares: 10
+		UseCashTick: yes
+	RevealMapCrateAction:
+		SelectionShares: 2
+		Effect: reveal-map
+	LevelUpCrateAction:
+		SelectionShares: 25
+	GiveMcvCrateAction@soviets:
+		SelectionShares: 0
+		NoBaseSelectionShares: 150
+		ValidFactions: cuba, libya, iraq, russia
+		Units: smcv
+	GiveMcvCrateAction@allied:
+		SelectionShares: 0
+		NoBaseSelectionShares: 150
+		ValidFactions: america, germany, england, france, korea
+		Units: amcv
+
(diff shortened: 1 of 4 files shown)


sequences/misc.yaml
```diff
@@ -42,16 +42,33 @@ crate-effects: #Temp Null Some SHPs cannot be read by openra at this time, expec
...
+		UseTilesetExtension: true
+	land:
+		Offset: 0, -12
+		ShadowStart: 1
+	water: wcrate
+		ShadowStart: 1
+		Offset: 0, -12
+		UseTilesetExtension: true
+
+wcrate:
```
Mailaender: This sequence is unused.

rules/misc.yaml
```diff
@@ -16,6 +16,38 @@ waypoint:
 	WithSpriteBody:
 	QuantizeFacingsFromSequence:
 
+CRATE:
```
Phrohdoh: Why `CRATE` and not `crate`?

MustaphaTR: All other mods had `Crate:`. If you want to change it to `crate:`, i'll.

Phrohdoh: Actor definitions in RA2 have all been lowercase so let's go with `crate:`.

MustaphaTR:
Updated but can't squash commits right now.

MustaphaTR:
Squashed.

abcdefg30:

The crates not showing their water sequence is an upstream issue I'll look into.

abcdefg30:
I filed https://github.com/OpenRA/OpenRA/pull/10568 which makes this work in conjunction with #80.

sequences/misc.yaml
```diff
@@ -42,16 +42,28 @@ crate-effects: #Temp Null Some SHPs cannot be read by openra at this time, expec
...
+	levelup: veteran
 	speed: #speed
 
+crate:
+	idle:
+		Offset: 0, -12
+		UseTilesetExtension: true
+	land:
+		Offset: 0, -12
+		ShadowStart: 1
```
Mailaender: I think you are missing `UseTilesetExtension: true` here which causes the game to crash:

> Exception of type `System.InvalidOperationException`: Unit `crate` does not have a sequence named `land`

MustaphaTR: I have no idea why i forgot that. I'll open a PR to fix, or you can do a direct commit without PR.

Mailaender: It will only crash in conjunction with https://github.com/OpenRA/OpenRA/pull/10568 by the way.

MustaphaTR: Becuase it uses `idle:` without this pr.

Mailaender: Can you send us a fixup PR?

MustaphaTR: Already sent.

PR #45, opened 2016-01-16
Setup Travis CI checks

Mailaender:
To help keep this repository stable.

PR #46, opened 2016-01-16
20160116 Upgrade

Mailaender:
Fixes the outdated rulesets after https://github.com/OpenRA/OpenRA/pull/10371 and https://github.com/OpenRA/OpenRA/pull/10364.

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -183,7 +183,7 @@ garefn:
 		Actor: cmin
 		SpawnOffset: 4, 1
 		Facing: 160
-	WithDockingOverlay:
+	WithDockedOverlay:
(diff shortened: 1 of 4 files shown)


Mailaender:
Please ignore the checks. The configuration from https://github.com/OpenRA/ra2/pull/45 is missing here.

abcdefg30:
Could we also introduce a `RequiresMods:` entry now?

Mailaender:
That would be cumbersome at this stage. `make version` will change for every commit on OpenRA/OpenRA, the game won't start and the lint rules will fail.

Mailaender:
Rebased. This will fix the build again.

abcdefg30:
:+1:

Issue #37, opened 2016-01-15
Crash on click to music button in Extras menu


yaaaaa:
```
Red Alert 2 Mod at Version {DEV_VERSION}
on map ba240d57ac307937bab3223ce71e0e37b84e3a9e (Blank by Matthias Mailänder).
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.42000
Exception of type `System.Reflection.TargetInvocationException`: Адресат вызова создал исключение.
Inner Exception of type `System.InvalidOperationException`: Widget MUSIC_PANEL has no child MUTE_LABEL of type LabelWidget
     в OpenRA.Widgets.Widget.Get[T](String id) в d:\Games\OpenRA-bleed\OpenRA.Game\Widgets\Widget.cs:строка 528
   в OpenRA.Mods.Common.Widgets.Logic.MusicPlayerLogic..ctor(Widget widget, Ruleset modRules, World world, Action onExit)
   в System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   в System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   в System.Reflection.ConstructorInfo.Invoke(Object[] parameters)
   в OpenRA.ObjectCreator.CreateUsingArgs(ConstructorInfo ctor, Dictionary`2 args) в d:\Games\OpenRA-bleed\OpenRA.Game\ObjectCreator.cs:строка 88
   в OpenRA.ObjectCreator.CreateObject[T](String className, Dictionary`2 args) в d:\Games\OpenRA-bleed\OpenRA.Game\ObjectCreator.cs:строка 53
   в OpenRA.Widgets.Widget.<>c__DisplayClass3.<PostInit>b__2(String l) в d:\Games\OpenRA-bleed\OpenRA.Game\Widgets\Widget.cs:строка 254
   в System.Linq.Enumerable.WhereSelectArrayIterator`2.MoveNext()
   в System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   в System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   в OpenRA.Widgets.Widget.PostInit(WidgetArgs args) в d:\Games\OpenRA-bleed\OpenRA.Game\Widgets\Widget.cs:строка 254
[...]

PR #47, opened 2016-01-17
Removed the bit rotted music player chrome

Mailaender:
Music installation is now handled by the mod chooser. Fixes https://github.com/OpenRA/ra2/issues/37.


abcdefg30:
Doesn't crash when selecting `[Install Music]`. :white_check_mark:

PR #49, opened 2016-01-17
Fixed soviet engineer speaking with allied voice set

Mailaender:
Closes https://github.com/Phrohdoh/OpenRA/issues/82.

### DIFF
--- a/audio/voices.yaml
+++ b/audio/voices.yaml
@@ -1,28 +1,26 @@
-AlliedEngineerVoice:
+EngineerVoice:
 	DefaultVariant: .wav
+	Prefixes:
+		america: iena
+		germany: iena
+		england: iena
+		france: iena
+		korea: iens
+		cuba: iens
+		libya: iens
+		iraq: iens
+		russia: iens
+	DisablePrefixes: Zapped, Burned, Melted
 	Voices:
-		Select: ienasea, ienaseb, ienasec, ienased
-		Move: ienamoa, ienamob, ienamoc
-		Action: ienaata, ienaatb, ienaatc
-		Feedback: ienafea, ienafeb, ienafec
-		Die: ienadia, ienadib, ienadic, ienadid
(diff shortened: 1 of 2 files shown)


abcdefg30:
Sorry, we can't take this at the moment: https://github.com/OpenRA/ra2/pull/1#discussion_r48202377

TL;DR: There is no `ienssed`, `iensatc` and `iensdid`, IIRC.

Mailaender:
That shouldn't be a blocker. If the engine can't find the file, it will only log that. Nothing serious happens.

abcdefg30:
But that would mean reverting this after we got an upstream fix like https://github.com/OpenRA/ra2/pull/1#discussion_r48284138:
```
Voiced:
    VoiceSet: AlliedEngineer
    FactionVoiceSets:
        allies: AlliedEngineer
        soviets: SovietEngineer
```
Or like what https://github.com/OpenRA/OpenRA/pull/10422 proposed.

Mailaender:
You probably mean https://github.com/OpenRA/OpenRA/pull/10050 which was already rejected.

GraionDilach:
>If the engine can't find the file, it will only log that. Nothing serious happens.

A muted response is IMO serious. But I guess I'm alone with this.

abcdefg30:
> A muted response is IMO serious. But I guess I'm alone with this.

I think it's a bug that needs fixing too, this at the moment just looks like trading bad yaml with bugs (although it was in principle a good idea to do this).

> You probably mean OpenRA/OpenRA#10050 which was already rejected.

I wouldn't say rejected, the author just closed his pr again (even so fast that I didn't recognized what's actually going on there).

GraionDilach:
This was the reason behind that PR's closure. http://logs.openra.net/?year=2015&month=11&day=21#11:36:32

Mailaender:
Superseded by https://github.com/OpenRA/ra2/pull/106.

PR #51, opened 2016-01-17
Added support for US "The First Decade" DVDs

Mailaender:
I don't own them so I can't test this.


abcdefg30:
@jaZzKCS reported that this isn't working: `[13:14:54] <jaZz_KCS> it let me select Install from CD, albeit errors instantly ---> "Installation failed: File not found: THEME.mix"`

Mailaender:
I made a new attempt https://github.com/OpenRA/ra2/pull/279.

PR #52, opened 2016-01-17
Add the ChronoshiftPower

abcdefg30:
Thanks to @tomsons26 for the basic work on this.

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -760,6 +760,17 @@ gacsph:
 		Sequence: idle-dome
 	RequiresPower:
 	CanPowerDown:
+	ChronoshiftPower@chronoshift:
+		OrderName: Chronoshift
+		Icon: chrono
+		ChargeTime: 120
+		Description: Chronoshift
+		LongDesc: Teleports a group of units across\nthe map.
+		# Those sounds don't work: OpenRA/OpenRA#10540
+		# They also want to be played successive
+		#SelectTargetSound: schrlo1.wav, schrlo2a.wav schrlo2b.wav, schrlo2c.wav, schrlo3.wav
+		EndChargeSound: schropen.wav
+		DisplayRadarPing: True
 
 atesla:
 	Inherits: ^SupportBuilding
--- a/cursors.yaml
+++ b/cursors.yaml
@@ -192,7 +192,10 @@ Cursors:
 			Length: 5
 		looped-patrol:
 			Start: 356
-		chronoshift:
+		chrono-select:
+			Start: 357
+			Length: 12
+		chrono-target:
 			Start: 357
 			Length: 12
 		remove-bomb:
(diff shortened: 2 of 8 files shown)


rules/allied-structures.yaml
```diff
@@ -760,6 +760,17 @@ gacsph:
...
+	ChronoshiftPower@chronoshift:
+		OrderName: Chronoshift
+		Icon: chrono
+		ChargeTime: 120
+		Description: Chronoshift
+		LongDesc: Teleports a group of units across\nthe map.
+		# Those sounds don't work: ChunkSize is 0
+		#SelectTargetSound: schrlo1.wav, schrlo2a.wav schrlo2b.wav, schrlo2c.wav, schrlo3.wav
+		EndChargeSound: schropen.wav
+		KillCargo: yes
```
GraionDilach: Wrong!

GraionDilach:
Notsure if default values trigger it, but RA2 Chronosphere doesn't bring back the units after a time and it only kills directly targeted infantry (excluding CLeg), not passengers.

abcdefg30:
> Notsure if default values trigger it, but RA2 Chronosphere doesn't bring back the units after a time 

I know, see https://github.com/OpenRA/ra2/pull/52/files#diff-ffa2589dfb42cb2d2ecffa68264dd139R409.

rules/allied-structures.yaml
abcdefg30: Removed it.

abcdefg30:
Just had an idea. We can make use of the `ExplodeInstead` flag on `Chronoshiftable` (which just kills stuff and doesn't require an explosion) to add the correct behaviour for infantry. They don't show a death animation yet though.

GraionDilach:
That sounds good to me as a start.

GraionDilach:
Oh, I see they have been added already. :+1: then but wait with this for someone else.

rules/allied-structures.yaml
```diff
@@ -760,6 +760,16 @@ gacsph:
...
 	RequiresPower:
 	CanPowerDown:
+	ChronoshiftPower@chronoshift:
+		OrderName: Chronoshift
+		Icon: chrono
+		ChargeTime: 120
+		Description: Chronoshift
+		LongDesc: Teleports a group of units across\nthe map.
+		# Those sounds don't work: ChunkSize is 0
+		#SelectTargetSound: schrlo1.wav, schrlo2a.wav schrlo2b.wav, schrlo2c.wav, schrlo3.wav
```
Mailaender: Filed https://github.com/OpenRA/OpenRA/issues/10540.

Mailaender:
I submitted https://github.com/abcdefg30/ra2/pull/1 to fix the crash I introduced recently.

> Exception of type `System.InvalidOperationException`: Unit `overlay` does not have a sequence named `target-valid-temperate`

abcdefg30:
Updated.

rules/allied-structures.yaml
abcdefg30: Actually, the default here is already `true`. Fixed that in #70.

PR #61, opened 2016-01-19
Added the Tiberian Sun cameo palette

Mailaender:
This is helpful when digging through the left-overs in the asset browser.


GraionDilach:
In this case, I'd add the TS anim.pal as well into this - s_clsn42 uses that.

MustaphaTR:
Power down anim also uses anim.pal iirc, but maybe palette.pal not sure. Adding that would be good too.

GraionDilach:
Palette.pal is shared between the two being the RA1 temperat.pal.

MustaphaTR:
I looked to Ares.dll Documentation, it says togle power uses mousepal.pal.

Mailaender:
I filed https://github.com/OpenRA/ra2/pull/63 for the TS mousepal.pal fix.

abcdefg30:
Looks good to me.

PR #66, opened 2016-01-20
Further polish to the powerdown indicator

abcdefg30:
The blocked animation looked weird (too small). I could use the green color instead of the light blue if wished. The next thing is that the powerdown indicator was below the offline building.
ra2powerdown

### DIFF
--- a/cursors.yaml
+++ b/cursors.yaml
@@ -273,7 +273,7 @@ Cursors:
 			Start: 73
 			Length: 5
 		powerdown-blocked:
-			Start: 52
+			Start: 342
 		joystick-all:
 			Start: 385
 		joystick-t-blocked:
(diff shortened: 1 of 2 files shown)


cursors.yaml
```diff
@@ -273,7 +273,7 @@ Cursors:
 			Start: 73
 			Length: 5
 		powerdown-blocked:
-			Start: 52
+			Start: 345
```
Mailaender: I think it may be worse repeating `339` here without animation, or `342` to show a greyed out version. This looks like an unrelated movement block cursor which is a bit irritating. Feels like I am not in powerdown mode.

abcdefg30:
Updated. Using the sequence index `342` now.

PR #67, opened 2016-01-20
Added the Dolphin

Mailaender:
http://cnc.wikia.com/wiki/Dolphin_(Red_Alert_2)

### DIFF
--- a/rules/allied-naval.yaml
+++ b/rules/allied-naval.yaml
@@ -129,3 +129,79 @@ aegis:
 	Voiced:
 		VoiceSet: AlliedNavalVoice
 
+dlph:
+	Inherits@1: ^ExistsInWorld
+	Inherits@2: ^SpriteActor
+	Inherits@3: ^GainsExperience
+	Buildable:
+		Queue: Ship
+		BuildPaletteOrder: 30
+		Prerequisites: gatech, ~gayard
+	Valued:
+		Cost: 500
+	Tooltip:
+		Name: Dolphin
+		Description: Trained dolphin\narmed with sonic beams.\n  Strong vs Ships
+	Armor:
+		Type: Light
+	Health:
+		HP: 200
+	Mobile:
+		ROT: 6
+		Speed: 90
+		TerrainSpeeds:
+			Water: 100
+		Voice: Move
+	SelectionDecorations:
+		Palette: pips
+	Selectable:
+	Voiced:
+		VoiceSet: DolphinVoice
+	Armament:
+		Weapon: SonicZap
+	AttackFrontal:
+		Voice: Attack
+	AutoTarget:
+	RevealsShroud:
+		Range: 4c0
+	Cloak:
+		CloakTypes: Underwater
+		InitialDelay: 50
+		CloakDelay: 75
+		CloakSound: vnavupa.wav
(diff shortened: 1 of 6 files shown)


rules/allied-naval.yaml
```diff
@@ -129,3 +129,85 @@ aegis:
...
+	Cloak:
+		CloakTypes: Underwater
+		InitialDelay: 50
+		CloakDelay: 75
+		CloakSound: vnavupa.wav
+		UncloakSound: vnavupa.wav
+		WhileCloakedUpgrades: underwater
+		UncloakOnAttack: false
+	Targetable:
+		TargetTypes: Ground, Water, Repair
```
abcdefg30: Why `Repair` here (and below)?

rules/allied-naval.yaml
```diff
@@ -129,3 +129,85 @@ aegis:
...
+	DetectCloaked:
+		CloakTypes: Underwater
+		Range: 4c0
+	HiddenUnderFog:
+	ActorLostNotification:
+	Guard:
+		Voice: Attack
+	Guardable:
+	DeathSounds@normal:
+		DeathTypes: ExplosionDeath, BulletDeath, FlameDeath, ElectroDeath, RadiationDeath
```
abcdefg30: You shouldn't need to define all that. If `DeathTypes` is empty a sound will always play: https://github.com/OpenRA/OpenRA/pull/10156

rules/allied-naval.yaml
Mailaender: Copy pasta. Removed it.

rules/allied-naval.yaml
Mailaender: Good to know. I had a long fight with the linter about this.

weapons.yaml
```diff
@@ -783,6 +783,42 @@ SubTorpedo:
...
+		AffectsParent: false
+		ValidStances: Neutral, Enemy
+		Versus:
+			Heavy: 80
+			Concrete: 60
+		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
+	Warhead@2Dam: SpreadDamage
+		Range: 0, 32
+		Falloff: 50, 50
+		Damage: 10
```
GraionDilach: The other warhead has 10 damage. Decide.

Mailaender:
Optimized the weapon a bit.

abcdefg30:
Ok, just some (minor) things:
- The weapon animation doesn't look right, but I guess we can't cut the trail at the moment?
- It uncloaks to fire, that should not happen
- What about using `NavalUnitLost` as `ActorLostNotification`?

Looks good to me otherwise.

Mailaender:
Yeah, I think this needs some adjustments to the laser rendering to make it cut the trail and gain more transparency. The game ignoring `UncloakOnAttack: false` is probably an upstream engine bug. Nothing I can do about these here. 

Added `NavalUnitLost`.

Mailaender:
Took https://github.com/OpenRA/OpenRA/pull/9808 into account to fix the uncloak problem and also slowed down the Soviet naval yard animations.

abcdefg30:
Looks good to me. Thanks. :+1:

PR #70, opened 2016-01-21
Fix the chronoshift killing passengers

abcdefg30:
Fixes an oversight in #52 , the value is actually true [by default](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.RA/Traits/SupportPowers/ChronoshiftPower.cs#L37).


MustaphaTR:
:+1:

PR #74, opened 2016-01-21
Remove the duplicated chronosphere cursor sequence

abcdefg30:
Depends on https://github.com/OpenRA/OpenRA/pull/10558.


abcdefg30:
OpenRA/OpenRA#10558 was merged.

PR #78, opened 2016-01-23
Added IFV passenger dependent turrets

Mailaender:
A prime example of https://github.com/OpenRA/OpenRA/wiki/Actor-Upgrades

Depends on https://github.com/OpenRA/OpenRA/pull/10567 for the Travis check to pass.

### DIFF
--- a/sequences/voxels.yaml
+++ b/sequences/voxels.yaml
@@ -65,9 +65,12 @@ naflak:
 flata: #Unused
 	idle:
 
-fv: #Has different turrets per unit thats inside,default is patriot
+fv:
 	idle:
-	turret: fvtur
+	turret-missile: fvtur
+	turret-mg: fvtur1
+	turret-repair: fvtur2
+	turret-hitech: fvtur3
 
 gtgcan:
 	turret: gtgcantur
--- a/rules/allied-infantry.yaml
+++ b/rules/allied-infantry.yaml
@@ -17,6 +17,7 @@ engineer:
 		Speed: 56
 	Passenger:
 		PipType: Blue
+		GrantUpgrades: ifv-repair
 	EngineerRepair:
 	RepairsBridges:
 	Captures:
@@ -99,6 +100,7 @@ e1:
 		Speed: 31
 	Passenger:
 		PipType: Green
+		GrantUpgrades: ifv-machinegun
 	RevealsShroud:
 		Range: 5c0
 	AttackFrontal:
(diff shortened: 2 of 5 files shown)


GraionDilach:
I'm not a fan of this method but indeed, the Tick Tank PR didn't got merged (which would have allowed `AttackTurreted` be upgraded instead of `Armament`s to allow eliteweapons on the additional IFV modes).

sequences/voxels.yaml
```diff
@@ -65,9 +65,21 @@ naflak:
...
+	turret-advm: fvtur3
+	turret-sniper: fvtur4
+	turret-chrono: fvtur5
+	turret-flak: fvtur6
+	turret-tesla: fvtur7
+	turret-demo: fvtur8
+	turret-deso: fvtur9
+	turret-psi: fvtur10
+	turret-hero: fvtur11
+	turret-engi: fvtur12
```
MustaphaTR: These are all wrong there are only 4 turret that really used, Missile, MG, Repair and Hi-Tech. Which are `fvtur`, `fvtur1`, `fvtur2` and `fvtur3`. Others are unused.

GraionDilach: The Cow Turret was used. IIRC the rest are duplicate files so no harm done.

MustaphaTR: Having them here doesn't harm but they are also defined like that under `WithVoxelTurret:`.

Mailaender: That fixes the glitches. Thanks guys.

abcdefg30:
OpenRA/OpenRA#10567 was merged.

Mailaender:
Rebased.

Iamgoofball:
Doesn't the IFV also get improved move speed if you stuff a dog in there?

GraionDilach:
RA2 never had such a logic at all.

abcdefg30:
Looks good to me. In the original other units (like the spy) gave the IFV a `machinegun` turret, too. But I'm not certain anymore and will recheck that. In addition this pr is already good enough, so :+1: / :white_check_mark:

Iamgoofball:
Mighta been YR then, idk, I swear dogs in IFVs would make them go faster

GraionDilach:
Don't swear. Such logic was impossible to do in the engine. Passengers couldn't affect stats of their cargo at all. The IFV was one giant hackish approach in WW and no, it only could change turret and weapon, never speed or any other stat, neither in RA2 neither in YR.

PR #80, opened 2016-01-23
Add missing IsWater entries to all tilesets

abcdefg30:
Noticed during the review of #40.

Note: It doesn't fix our "crates don't display their water animation" issue.


MustaphaTR:
I've already noticed and tried this to fix water issue of crates. I know this doesn't fixes it. :+1:

PR #91, opened 2016-01-24
Fix Spies can enter all Buildings

MustaphaTR:
Now they only enter what they supposed to enter except the SWs because something like `InfiltrateForSWTimerReset:` is not added yet iirc.


Mailaender:
You can also setup https://github.com/OpenRA/OpenRA/wiki/Traits#infiltrateforexploration on enemy radar facilities already.

Mailaender:
Nevermind, we already have that.

Mailaender:
Thanks!

PR #92, opened 2016-01-24
Added transparency for underwater units

Mailaender:
Red Alert 2 used pre-multiplied alpha for this not a fixed palette like RA95.


abcdefg30:
Nice, thanks. :+1:

PR #77, opened 2016-01-23
Add all speech notifications to notifications.yaml

abcdefg30:
And also make use of them. Followups are on their way.

Depends on https://github.com/OpenRA/OpenRA/pull/10564 at the moment.

PR #93, opened 2016-01-24
Added the Giant Squid

Mailaender:
without `SquidGrab` only `SquidPunch`. See http://cnc.wikia.com/wiki/Giant_squid and https://gist.github.com/Mailaender/c2c5ff37a8927d0bc452 for reference. Also fixes https://github.com/OpenRA/ra2/issues/132.

### DIFF
--- a/rules/soviet-naval.yaml
+++ b/rules/soviet-naval.yaml
@@ -207,3 +207,6 @@ sqd:
 	WithInfantryBody:
 		StandSequences: stand
 		AttackSequence: punch
+	Repairable:
+		Voice: Attack
+		RepairBuildings: gayard, nayard
--- a/rules/soviet-naval.yaml
+++ b/rules/soviet-naval.yaml
@@ -135,3 +135,75 @@ hyd:
 		EmptyWeapon: UnitExplode
 	WithVoxelBody:
 	RenderSprites:
+
+sqd:
+	Inherits@1: ^ExistsInWorld
+	Inherits@2: ^SpriteActor
+	Inherits@3: ^GainsExperience
+	Buildable:
+		Queue: Ship
+		BuildPaletteOrder: 30
+		Prerequisites: natech, ~nayard
+	Valued:
+		Cost: 500
+	Tooltip:
+		Name: Giant Squid
+		Description: Ocean creature\npunches enemies in close combat.\n  Strong vs Ships
+	Armor:
+		Type: Light
+	Health:
+		HP: 200
+	SelfHealing:
+	Mobile:
+		ROT: 6
+		Speed: 90
+		TerrainSpeeds:
+			Water: 100
+		Voice: Move
+	SelectionDecorations:
+		Palette: pips
+	Selectable:
+	Voiced:
+		VoiceSet: GiantSquidVoice
+	Armament:
+		Weapon: SquidPunch
+	AttackFrontal:
+		Voice: Attack
+	AutoTarget:
+	RevealsShroud:
+		Range: 5c0
+	Cloak:
+		CloakTypes: Underwater
+		InitialDelay: 50
+		CloakDelay: 75
+		CloakSound: vnavupa.wav
+		UncloakSound: vnavupa.wav
+		WhileCloakedUpgrades: underwater
+		UncloakOn: Damage
+		IsPlayerPalette: true
+	Targetable:
+		TargetTypes: Ground, Water
+		UpgradeTypes: underwater
+		UpgradeMaxEnabledLevel: 0
+	Targetable@underwater:
+		TargetTypes: Underwater
+		UpgradeTypes: underwater
+		UpgradeMinEnabledLevel: 1
+		IsPlayerPalette: true
+	DetectCloaked:
+		CloakTypes: Underwater
+		Range: 5c0
+	HiddenUnderFog:
+	ActorLostNotification:
+	Guard:
+		Voice: Attack
+	Guardable:
+	DeathSounds:
+	QuantizeFacingsFromSequence:
+		Sequence: stand
+	Chronoshiftable:
+		ExplodeInstead: true
+	RenderDetectionCircle:
+	WithInfantryBody:
+		StandSequences: stand
+		AttackSequence: punch
--- a/weapons.yaml
+++ b/weapons.yaml
@@ -874,6 +874,7 @@ SonicZap:
 	ReloadDelay: 180
 	Range: 6c0
 	Report: vdolatta.wav
+	ValidTargets: Water, Underwater
 	Burst: 2
 	Projectile: AreaBeam
 		Speed: 1c0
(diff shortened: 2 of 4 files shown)


rules/soviet-naval.yaml
```diff
@@ -132,3 +132,81 @@ hyd:
...
+	DeathSounds:
+	MustBeDestroyed:
+		RequiredForShortGame: true
+	QuantizeFacingsFromSequence:
+		Sequence: stand
+	Chronoshiftable:
+		ExplodeInstead: true
+	RenderDetectionCircle:
+	RenderSprites:
+	WithInfantryBody:
```
GraionDilach: This should IMO use WithFacingSpriteBody instead (Squid animated played the move anim on a slower rate while standing still and IIRC WFSB can pull that off).

Mailaender: How do I setup `AttackSequence` using `WithFacingSpriteBody`?

GraionDilach: Use `WithAttackAnimation`.

Mailaender: Tried it. `WithFacingSpriteBody` won't play an animation during movement.

rules/soviet-naval.yaml
```diff
@@ -132,3 +132,81 @@ hyd:
...
+		Range: 5c0
+	HiddenUnderFog:
+	ActorLostNotification:
+		Notification: NavalUnitLost
+	Guard:
+		Voice: Attack
+	Guardable:
+	DeathSounds:
+	MustBeDestroyed:
+		RequiredForShortGame: true
```
abcdefg30: This doesn't look correct.

rules/soviet-naval.yaml
```diff
@@ -132,3 +132,81 @@ hyd:
...
+	QuantizeFacingsFromSequence:
+		Sequence: stand
+	Chronoshiftable:
+		ExplodeInstead: true
+	RenderDetectionCircle:
+	RenderSprites:
```
abcdefg30: This should be already inherited from `^SpriteActor`.

rules/soviet-naval.yaml
```diff
@@ -132,3 +132,81 @@ hyd:
...
+	WithInfantryBody:
+		StandSequences: stand
+		AttackSequence: punch
+	WithDeathAnimation:
+		UseDeathTypeSuffix: false
```
abcdefg30: I somehow think we can drop that. The squids didn't have death animations and the weapon explosion should hide that anyway.

GraionDilach: What explosion? Squids and Dolphins didn1t explode at all.

abcdefg30: Weapon (impact) causing its death.

GraionDilach: The animals have a death animation though. This is completely legit.

abcdefg30: I don't remember one. Anyway, do you know where we can find the squids death animation then?

MustaphaTR: Squids don't have a death anim. Dolphins have a unused jumping anim but that is possibly cheer, not death.

GraionDilach: Hm, now I rechecked. Apparently I remembered wrong and they do lack death anims - the YR dinos got such which probably plays a role why I remembered these also got such I guess.

MustaphaTR: Dinos are infantry types. Squird amd dolphins are vehicle.

Mailaender:
Updated.

rules/soviet-naval.yaml
```diff
@@ -132,3 +132,76 @@ hyd:
...
+		TargetTypes: Underwater
+		UpgradeTypes: underwater
+		UpgradeMinEnabledLevel: 1
+		IsPlayerPalette: true
+	DetectCloaked:
+		CloakTypes: Underwater
+		Range: 5c0
+	HiddenUnderFog:
+	ActorLostNotification:
+		Notification: NavalUnitLost
```
abcdefg30: This isn't correct anymore since #77 was merged.

abcdefg30:
Some other things I noticed:
- The self heal is a bit too fast
- It doesn't do damaged to dolphins/subs (but dolphins don't do damage to it, too)
- Weren't both squids and dolphins repairable (at the naval yards)?

Looks good to me otherwise.

MustaphaTR:
> Weren't both squids and dolphins repairable (at the naval yards)?

Yes, they were.

Mailaender:
Updated. Also fixed the weapon sounds. I always miss the .wav file extension.

abcdefg30:
Somehow squids still don't do any damage.

You can also still damage land units with dolphins, but I don't have in mind if this is fixable yet:
dolphinlandattack

MustaphaTR:
> You can also still damage land units with dolphins, but I don't have in mind if this is fixable yet:

This happens in  orginal RA2 too if you target them from a short piece of land that they can reach to other side.

rules/soviet-naval.yaml
Mailaender: I currently use the jumping as idle which is a bit weird as it doesn't resurface. Cheer sounds like a better plan. That is used when the player won?

MustaphaTR: Cheer is used in missions by a action numbered as 113 (iirc). Also you can do cheer by clicking C or whatever you set it. There is also a unused but enablable by editing cheer Advaned Command Bar button for Cheer.

Mailaender:
It is fixable, though not here, but at https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Effects/AreaBeam.cs

GraionDilach:
Needs a rebase. Also please set `ValidTargets` on the Dolphin warheads as well (not just on the weapon), that will degrade the bug to purely visuals.

GraionDilach:
Oh, `SquidPunch` also needs the `ValidTargets` defined on the warhead as well.

Mailaender:
Rebased and added more `ValidTargets`.

weapons.yaml
```diff
@@ -905,6 +907,20 @@ SonicZap:
...
 		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
+		ValidTargets: Water, Underwater
+
+SquidPunch:
+	ReloadDelay: 32
+	Range: 1c856
+	Report: vsquat1a.wav
+	ValidTargets: Water, Underwater
+	Projectile: Bullet
+		Speed: 30
```
abcdefg30: This adds a weird delay in between the animation and the impact.
How about using a higher value (maybe `3000`, don't know if that's already overkill, though)?

abcdefg30:
Looks good to me otherwise. :+1: / :white_check_mark: after above is fixed.

weapons.yaml
Mailaender: I believe 100 is instant or at least normal bullet speed. Updated.

GraionDilach: 100c0 is instant, not 100.

abcdefg30:
Thanks!

PR #98, opened 2016-01-31
Fixed case-sensitivity errors on tileset sprites

Mailaender:
See https://github.com/OpenRA/OpenRA/pull/10531.


abcdefg30:
:+1: thanks.

PR #100, opened 2016-01-31
Fixed harvesters spawning too far from their refinery

Mailaender:
as done for Tiberian Sun in https://github.com/OpenRA/OpenRA/pull/10573.


abcdefg30:
Looks good to me. :+1:

PR #102, opened 2016-02-01
Fix the package order in mod.yaml and remove the Folders entry

abcdefg30:
The required fixup after https://github.com/OpenRA/OpenRA/pull/10554 is merged.

### DIFF
--- a/mod.yaml
+++ b/mod.yaml
@@ -7,24 +7,10 @@ Metadata:
 RequiresMods:
 	modchooser: {DEV_VERSION}
 
-Folders:
+Packages:
+	~^Content/ra2
 	.
 	ra2:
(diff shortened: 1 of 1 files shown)


mod.yaml
```diff
@@ -7,7 +7,8 @@ Metadata:
 RequiresMods:
 	modchooser: {DEV_VERSION}
 
-Folders:
+Packages:
+	~^Content/ra2
 	.
 	ra2:
 	ra2:bits
```
pchote: You will want to define these after the content mix files if you want to be able to override any of the mix file contents.

abcdefg30: Yeah, updated.

PR #104, opened 2016-02-02
Added the MCV transform sound

Mailaender:
https://gist.github.com/Mailaender/c2c5ff37a8927d0bc452#file-sound-ini-L1999


GraionDilach:
This is wrong! :-1:

MoveStarts are played when the unit starts to move and they are actual motor/engine sounds.

Mailaender:
We don't have movement sounds yet.

Mailaender:
@pchote You were wrong btw.

pchote:
?

Mailaender:
Misunderstanding. Filed https://github.com/OpenRA/OpenRA/issues/10685 and https://github.com/OpenRA/ra2/pull/105.

PR #105, opened 2016-02-03
Fixed Allied MCV and Chrono Miner voices

Mailaender:
https://gist.github.com/Mailaender/c2c5ff37a8927d0bc452#file-sound-ini-L1642

### DIFF
--- a/audio/voices.yaml
+++ b/audio/voices.yaml
@@ -31,6 +31,20 @@ AlliedVehicleVoice:
 		Move: vgramoa, vgramob, vgramoc, vgramod, vgramoe, vgramof
 		Attack: vgraata, vgraatb, vgraatc, vgraatd, vgraate
 
+AlliedConstructionVehicleVoice:
+	DefaultVariant: .wav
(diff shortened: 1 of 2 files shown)


GraionDilach:
What is the point of this? Most vehicles had a shared voice in RA2, YR added separate vehicle voices.

Mailaender:
Apparently MCV and chrono miner already had some move sounds.

MustaphaTR:
Move sound i think should go to its own trait. They are sound, not voice.

Mailaender:
But this is about the voice, not the movement sound.

GraionDilach:
Yes, which is still the same, so comes back to my initial question: what's the point of this?

Mailaender:
It corrects the allied MCV and chrono miner voice sets which aren't generic (at least for Move).

GraionDilach:
Ah, moa and moc aren't used there. Dammit, took me a while. :+1: then.

abcdefg30:
:+1:

PR #108, opened 2016-02-04
Added sprite sequence checks

Mailaender:
also with https://github.com/OpenRA/OpenRA/pull/10687 in mind. ~~Depends on https://github.com/travis-ci/apt-package-whitelist/issues/2380.~~ (Actually not as p7zip-full is already allowed.)


abcdefg30:
Looks ok to me and travis still passes. :+1:

Mailaender:
>  travis still passes

This is actually a bug. https://travis-ci.org/OpenRA/ra2/builds/107465160 should have failed. Filed a fix at https://github.com/OpenRA/OpenRA/pull/10802.

PR #109, opened 2016-02-05
New RA2 UI (soviets faction updated)

x-a-n-a-x:
New RA2 UI (soviets faction updated)


Mailaender:
Can we keep power down for now? It is useful for testing and debugging while the mod is in heavy development.

abcdefg30:
Doesn't this only remove the button, but not the feature itself?
If we want to keep it for debug reasons then we can still access it via the hotkeys.

x-a-n-a-x:
@abcdefg30 : indeed, I only modified the graphics.

Mailaender:
> If we want to keep it for debug reasons then we can still access it via the hotkeys.

The hotkey is tied to the UI widget.

PR #110, opened 2016-02-07
Fixed ingame-diplomacy.yaml not found crash

Mailaender:
as per https://github.com/OpenRA/OpenRA/pull/10650.


abcdefg30:
Just a note for all that wonder: This had/has my :+1:, I was just not at home and logged in into github.

PR #112, opened 2016-02-07
Fixed oil derrick and hospital not using tileset variants

Mailaender:
Visible on snow maps.


Mailaender:
Superseded by https://github.com/OpenRA/ra2/pull/114.

Issue #48, opened 2016-01-17
Missing prop actors [$6]


Mailaender:
Just a collection of what I stumbled upon during map import experiments:

- [ ] `camex02`
- [ ] `camex03`
- [ ] `camex05`

- [ ] `camsc07`
- [ ] `camsc08`
- [ ] `camsc09`

- [ ] `cawt01`
- [ ] `cahse01`
- [ ] `cafrma`
- [ ] `camsc10`
- [ ] `cawt01`
- [ ] `canewy20`
- [ ] `canewy21`
- [ ] `ptruck`
- [ ] `pole01`

- [ ] `cabunk01`
- [ ] `castrt05`
- [ ] `caeuro05`
- [ ] `capark01`
- [ ] `capark02`
- [ ] `capark03`
- [ ] `caeuro05`

- [ ] `cahse02`
- [ ] `cahse03`
- [ ] `cahse04`
- [ ] `cahse07`

- [ ] `camsc01`
- [ ] `camsc10`
- [ ] `camsc10`
- [ ] `cawash06`
- [ ] `capark03`
- [ ] `capark02`
- [ ] `capars13`
- [ ] `capars13`
- [ ] `caeur1`
- [ ] `caeur1`
- [ ] `caeur1`
- [ ] `caeur2`
- [ ] `caeur2`
- [ ] `caeur2`
- [ ] `camsc02`
- [ ] `camsc03`
- [ ] `camsc02`
- [ ] `camsc03`
- [ ] `cafrmb`
- [ ] `cafrmb`
- [ ] `cawash06`
- [ ] `capark02`
- [ ] `capark03`
- [ ] `camov01`
- [ ] `camov02`
- [ ] `camov02`
- [ ] `camov01`
- [ ] `catexs06`
- [ ] `catexs07`
- [ ] `catexs07`
- [ ] `catexs08`
- [ ] `canewy08`
- [ ] `canewy06`
- [ ] `canewy13`
- [ ] `canewy15`
- [ ] `cathosp`
- [ ] `cathosp`
- [ ] `caairp`
- [ ] `caairp`
- [ ] `canewy21`
- [ ] `canewy21`
- [ ] `canewy21`
- [ ] `canewy20`
- [ ] `canewy20`
- [ ] `canewy20`
- [ ] `canewy21`
- [ ] `canewy21`
- [ ] `canewy21`
- [ ] `canewy20`
- [ ] `canewy20`
- [ ] `canewy20`
- [ ] `carus05`
[...]

PR #114, opened 2016-02-07
Added missing civilian buildings and props

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/48.

The rules https://github.com/OpenRA/OpenRA/pull/10714 and sequences https://github.com/OpenRA/OpenRA/pull/10715 were automatically converted and then manually cleaned to use inheritance.

I noticed a lot of wrongdoings in our current rule setup. Many were hard-coding one sprite variant which doesn't come with snow in the arctic tileset. Should be all fixed now. Also got rid of the crash when the repair crane gets damaged already detected by https://github.com/OpenRA/ra2/pull/108.

### DIFF
--- a/rules/civilian-props.yaml
+++ b/rules/civilian-props.yaml
@@ -102,29 +102,372 @@ camsc13:
 	Armor:
 		Type: Concrete
 
-calit01e:
+ammocrat:
 	Inherits: ^CivBuilding
 	Tooltip:
-		Name: Light
+		Name: Ammo Crates
+	Health:
+		HP: 1
+	RenderSprites:
+		Image: ammo01
+
+casin01e:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Roadblock
+	Armor:
+		Type: concrete
+	Health:
+		HP: 5
+
+casin01n:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Roadblock
+	Armor:
+		Type: concrete
+	Health:
+		HP: 5
+
+casin01s:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Roadblock
+	Armor:
+		Type: concrete
+	Health:
+		HP: 5
+
+casin01w:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Roadblock
+	Armor:
+		Type: concrete
+	Health:
+		HP: 5
+
+camsc01:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Hot Dog Stand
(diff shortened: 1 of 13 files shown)


rules/palettes.yaml
```diff
@@ -8,6 +8,21 @@
 		Tileset: TEMPERATE
 		Filename: unittem.pal
 		ShadowIndex: 1
+	PaletteFromFile@temperatecity:
+		Name: tsplayer
```
GraionDilach: wrong?

Mailaender: Yes, this was my first guess. Fixed.

rules/civilian-props.yaml
```diff
@@ -102,29 +102,467 @@ camsc13:
...
+casin01e:
+	Inherits: ^CivBuilding
+	Tooltip:
+		Name: Roadblock
 	Armor:
-		Type: Steel
+		Type: concrete
+	Health:
+		HP: 5
+	Building:
```
abcdefg30: This shouldn't be necessary as it is inherited from `^CivBuilding`which inherits this from `^Building`.
(Same for the actors below that inherit `^CivBuilding`.)

abcdefg30:
You can drop the `Armor: Type: wood` from all actors inheriting `^CivBuilding`, too.

I also think that the civilian vehicles would profit from a `^CivVehicle` definition which contains:
```
Inherits: ^Vehicle
Valued:
	Cost: 800
Mobile:
	Speed: 113
Health:
	HP: 100
Armor:
	Type: Light
RevealsShroud:
	Range: 5c0
Selectable:
	Bounds: 64, 52, 0, -6
RenderSprites:
RenderVoxels:
WithVoxelBody:
Voiced:
	VoiceSet: CarVoice
```

You can entirely drop `ROT: 5`, `RenderSprites:` and `RenderVoxels` as they are inherited, too.

I also saw some weird graphical glitches with the new buildings:
mckongglitch
It seems like that is the "husk" which should be left by the building (not necessarily in this pr ;)) which is rendered already for some reason (wrong `ShadowStart` definitions?).

I guess some time we also want to rework the tooltips of those `NY Building <number>` actors.

Anyway, sorry for causing inconvenience here, but I think our yaml will be much cleaner with the extra defaults. If you wish I can send a PR against your fork with the changes I want.

Other than that good job!

Mailaender:
Rebased, simplified rules and fixed some shadow sequences. Husks are indeed also already doable. Will do it in a followup as this is already pretty large.

abcdefg30:
Looks good to me now. :+1: / :white_check_mark: 
Thanks.

PR #120, opened 2016-02-08
Add the NukePower of namisl

abcdefg30:
`Gentleman, it's a nuclear device!`

The first commit adds a camera definition, which is mostly taken from ra. I also took the map editor image from there.

I tried to recreate the nuke with the original values. An exception here is the `Falloff`. It ends with `25, 0, 0` because the original had `PercentAtMax=.25`, but OpenRA works differently so I added two times zero at the end so the nuke doesn't damage actors half across the map.

Also the flying missile still uses the wrong palette, I already filed OpenRA/OpenRA#10723 to fix this.

The animations and not all sounds are added yet, as OpenRA still doesn't support certain things (e.g. already open the silo one minute before the nuke is ready).


weapons.yaml
```diff
@@ -1639,3 +1639,34 @@ NukePayload:
 		SmudgeType: LargeScorch
 		Size: 3
 		Delay: 15
+
+atomic:
```
MustaphaTR: In fact nuke weapon and nuke reactor explodsion both uses NukePayload weapon. I recommend you to edit it. It also should polish nuke reactor explosion.

abcdefg30:
Updated. Depends on #153 now.

abcdefg30:
#153 was merged.

Mailaender:
https://github.com/OpenRA/OpenRA/pull/10723/ was merged, but you forgot to add `MissilePalette: player` and I think that will also still break the remapping.

abcdefg30:
Ops, updated.

Mailaender:
Some obvious lack of polishing with the timing, remap and smudges, but it works. :boom:

PR #121, opened 2016-02-08
Fix the beacon look by using PlaceSimpleBeacon

abcdefg30:
Before:
beaconb
After:
beacona
It also doesn't hide behind actors anymore now (`ZOffset: 2535`).

### DIFF
--- a/sequences/misc.yaml
+++ b/sequences/misc.yaml
@@ -40,12 +40,10 @@ rallypoint:
 		ZOffset: -2047
 
 beacon:
-	arrow: mouse
-		Start: 6
-		Offset: 1, -12
-	circles: ring
-		Length: 12
-		BlendMode: Additive
+	idle: pbeacon
+		Length: 20
+		Tick: 200
+		ZOffset: 2047
 
 crate-effects: #Temp Null Some SHPs cannot be read by openra at this time, expect more
 	Defaults: null
--- a/rules/player.yaml
+++ b/rules/player.yaml
@@ -54,7 +54,9 @@ Player:
 	HarvesterAttackNotifier:
 		Notification: OreMinerUnderAttack
 	PlayerStatistics:
-	PlaceBeacon:
+	PlaceSimpleBeacon:
+		IsPlayerPalette: true
+		Palette: player
 	ProvidesTechPrerequisite@infonly:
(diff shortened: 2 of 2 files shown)


Mailaender:
I suggest removing the ring entirely. It looks glitchy.

MustaphaTR:
There are beacons in orginal game. Why are we using those. (Same goes to TS.)

Mailaender:
We haven't identified the sprite sequences. If you know them, a pull request would be nice.

GraionDilach:
I'm not certain if the orig games had a rally point beacon, the actual beacon coming to my mind is the MP beacon (sidec01.mix:pbeacon.shp, unittem.pal)

MustaphaTR:
I'm talking about same beacon with @GraionDilach. There was indeed no rally point beacon.

GraionDilach:
Well, I wouldn't advise using the multiplayer beacon for rally points then. I consider this makeshift artwork more appropriate.

MustaphaTR:
Isn't what @abcdefg30 currently changing MP Beacon. Rally point flag on TS/RA2 mod are different and iirc they are flares.

GraionDilach:
No, this is about the rally point beacon.

MustaphaTR:
This is MP Beacon. Look the changes. Rallypoint flag is one above.

sequences/misc.yaml
```diff
@@ -39,9 +39,11 @@ rallypoint:
 		BlendMode: Additive
 
 beacon:
+	Defaults:
+		ZOffset: 2535
 	arrow: mouse
```
Mailaender: I assume we want the remapable `pbeacon` here.

Mailaender:
You can use 

```yaml
beacon:
	idle: pbeacon
		Length: 20
		Tick: 200
  		ZOffset: 2047
```

with the following instead of `PlaceBeacon` in `player.yaml`

```yaml
PlaceSimpleBeacon:
	IsPlayerPalette: true
	Palette: player
```

if you wait for https://github.com/OpenRA/OpenRA/pull/10756/.

abcdefg30:
Updated. Depends on OpenRA/OpenRA#10756 now. Thanks @Mailaender.

PR #122, opened 2016-02-08
Fixed external services relying on MIX file installation

Mailaender:
Similar to https://github.com/OpenRA/OpenRA/pull/10722. See also http://resource.openra.net/maps/8866/


abcdefg30:
Looks good to me, thanks. :+1: / :white_check_mark:

PR #136, opened 2016-02-12
Add prone-shoot sequence to GI

### LINKED ISSUE #135: Define the GI shoot-prone sequence
The GI has prone firing frames but these are not defined in the sequence yaml.

Phrohdoh:
Closes #135.

### DIFF
--- a/sequences/allied-infantry.yaml
+++ b/sequences/allied-infantry.yaml
@@ -166,6 +166,11 @@ e1:
 		Length: 6
 		Facings: 8
 		Tick: 80
+	prone-shoot:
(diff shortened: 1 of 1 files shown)


Mailaender:
Still looks glitchy flickering from stand to prone when firing salves as `prone-stand` is still missing, but somehow an improvement.

Phrohdoh:
Ah so that is the issue, I couldn't figure out what was missing.
Those will probably be 260 and 261.

Mailaender:
Same problem with the Russian conscripts btw.

PR #138, opened 2016-02-14
Add prone-stand to e1 and e2 (gi and cons)

Phrohdoh:
Address the comments [here](https://github.com/OpenRA/ra2/pull/136#issuecomment-183755939).

### DIFF
--- a/sequences/allied-infantry.yaml
+++ b/sequences/allied-infantry.yaml
@@ -122,6 +122,11 @@ e1:
 	Defaults: gi
 	stand:
 		Facings: 8
+	prone-stand:
+		Start: 260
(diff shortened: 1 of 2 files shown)


sequences/allied-infantry.yaml
```diff
@@ -122,6 +122,11 @@ e1:
 	Defaults: gi
 	stand:
 		Facings: 8
+	prone-stand:
+		Start: 260
```
Mailaender: Could also be `86`?

Mailaender: Nevermind, that is already `prone-run`.

PR #139, opened 2016-02-14
Fixed rally point offsets and artwork

Mailaender:
The placeholders looked very strange.


abcdefg30:
Looks much better. :+1:

Issue #127, opened 2016-02-11
Slight overhang (Allied) War Factory [$5]


jtyle6:
As you can bellow, there is Grizzly's being covered by the Factory.

![allied war factory](https://cloud.githubusercontent.com/assets/12602985/12977124/92ed8ce4-d114-11e5-82c7-e8042d9b0d76.png)

Fincer:
This is a general issue with units & buildings which are next to (Allied) War Factory. They're rendered under the factory as seen in the picture above.

I doubt this may be related to the following OpenRA (game engine) issue:
https://github.com/OpenRA/OpenRA/issues/7520

Or maybe just a separate bug. Not sure.

jtyle6:
I didn't know, that bug so old.

GraionDilach:
That is a different z-sorting issue. In this case I'd moreso guess the z-buffer is either ain't working well with RA2 cellsizes either we're using arbitrary numbers as z offsets. Or both.

Fincer:
Okay. Thanks for clarification.

Just another pic demonstrating the issue:

![alliedww-ra2](https://cloud.githubusercontent.com/assets/3705231/12996031/ccae67d0-d134-11e5-9615-d3c494eed7aa.png)

Fincer:
I did a little testing with Soviet War Factory. Can't reproduce this issue with it.

pchote:
> In this case I'd moreso guess the z-buffer is either ain't working well with RA2 cellsizes either 

The z-buffer isn't working *at all*.  It doesn't exist yet in the OpenRA engine (see https://github.com/OpenRA/OpenRA/issues/7520).

Context for others: The draw order is currently determined per-object based on the actor CenterPosition's y coordinate.  This behaves badly for the isometric mods because the primary art axes run along the diagonals, which means that the plane of "x in front of y" varies with pixel position (as specified in `buildngz.shp`), and so the draw order testing needs to be done per-pixel in the sprite shader.  This is a straightforward problem to solve, and I have a prototype that does it.  It just hasn't been moved upstream yet.

Mailaender:
I lessened it in https://github.com/OpenRA/ra2/pull/145 by not rendering the overlays needlessly all the time.

PR #145, opened 2016-02-21
Fixed the war factory overlays

Mailaender:
Fixes https://github.com/OpenRA/ra2/issues/127 and the soviet war factory which was more severely broken.

~~Depends on https://github.com/OpenRA/OpenRA/pull/10805.~~


abcdefg30:
Looks much better, thanks! :+1: / :white_check_mark:

PR #153, opened 2016-03-09
Update the mod to use map format 9

abcdefg30:
Also let the utility remove the new lines it doesn't like. We should just avoid adding them again.


abcdefg30:
OpenRA\OpenRA#10839 just got merged and bumped the format to `10`. Will update this tomorrow.

pchote:
Some of the format 10 updater changes will be reverted, so you should hold until that gets sorted.

abcdefg30:
Updated.

abcdefg30:
Merging with the :+1: of @Phrohdoh.

PR #155, opened 2016-03-13
Fixed the Travis CI build

Mailaender:
It didn't matter, because it is just an empty dummy, but that will change soon.


abcdefg30:
https://travis-ci.org/OpenRA/ra2/builds/115675205#L1180 looks fine. :white_check_mark:

Issue #116, opened 2016-02-07
Civilians [$5]


Mailaender:
are still missing:

- [ ] [CIV1]
- [ ] [CIV2]
- [ ] [CIV3]

- [ ] [CIVA]
- [ ] [CIVB]
- [ ] [CIVC]

- [ ] [CIVBBP]
- [ ] [CIVBFM]
- [ ] [CIVBF]
- [ ] [CIVBTM]
- [ ] [CIVSFM]
- [ ] [CIVSF]
- [ ] [CIVSTM]

and special characters:

- [ ] [VLADIMIR] ; Vladimir
- [ ] [PENTGEN] ; Pentagon General
- [ ] [PRES] ; President
- [ ] [SSRV] ; Secret Service

PR #157, opened 2016-03-13
Added Civilians

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/116.

### DIFF
--- a/audio/voices.yaml
+++ b/audio/voices.yaml
@@ -357,7 +357,7 @@ CivilianAlliedMaleVoice:
 		Burned: ienadia #gotta put something as temp else game crashes
 		Melted: igenmela, igenmelb, igenmelc
 
-CivilianSovietMalevVoice:
+CivilianSovietMaleVoice:
 	DefaultVariant: .wav
 	Voices:
 		Select: icissea, icisseb, icissec, icissed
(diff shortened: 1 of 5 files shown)


GraionDilach:
You'll have to replace die3 onwards on all of them with the RA2 death effects from another infantry sequence (to give them rad, electro and whatnot deaths).

Maybe coulda even adopt a custom sequence template and inheriting that everywhere.

Mailaender:
Updated.

abcdefg30:
It would be nice if you could remove those `Start: 0`, `Length: 1` and `Facings: 1` junk values.
As you don't use `Defaults:` here they should be unnecessary (likewise the `Offset: 0,0` on the icons).

Mailaender:
Removed default values.

GraionDilach:
I noticed that `idle2` is empty at many of the sequences. Could you duplicate `idle1`/override `WithInfantryBody` to not use it?

Then this is ready to go I think.

Mailaender:
Removed the empty `idle2`. It was probably a left-over from auto-imports, that I forgot to clean up.

GraionDilach:
This broke Travis.

Mailaender:
Vladimir has an `idle2` sequence as the only civilian infantry and the `Facings: 8` probably mean my sequence importers still need work.

abcdefg30:

PR #160, opened 2016-03-13
Fixed tab completion for developer commands, editor actor loading and map editor copy & paste.

Mailaender:
https://github.com/OpenRA/OpenRA/pull/10804 and https://github.com/OpenRA/OpenRA/pull/10919 combined. We also forgot about https://github.com/OpenRA/OpenRA/pull/9885.


abcdefg30:
Thanks for the update! :+1:

PR #162, opened 2016-03-13
Added the Chrono Legionnaire

Mailaender:
It is bit in placeholder quality due to the exotic weaponry. ~~Not sure why I can't get it to shoot at all though.~~


abcdefg30:
We're still missing: `ProducibleWithLevel` and the offset of the selection bounds (?) is wrong.

weapons/allied-infantry.yaml
```diff
@@ -294,3 +294,37 @@ awp:
...
+		Duration: 10
+		DamageInterval: 5
+		Width: 0c512
+		Shape: Flat
+		ZOffset: 2047
+		BeyondTargetRange: 0c256
+		Blockable: true
+		Color: 00FFFFFF
+		ValidTargets: Ground
+		Versus:
```
MustaphaTR: Why versus vaule is on Projictle?

That's possibly why it doesn't fire.

Mailaender:
Updated. It now fires and has has selection bounds. Thanks guys.

weapons/allied-infantry.yaml
```diff
@@ -294,3 +294,33 @@ awp:
...
+	Report: ichratta.wav
+	Projectile: LaserZap
+		BeamWidth: 2
+		BeamDuration: 7
+		Color: 00FFFFFF
+	Warhead@Damage: SpreadDamage
+		Spread: 64
+		Damage: 8
+		Versus:
+			None: 100
```
GraionDilach: You don't need these, 100 is default.

weapons/allied-infantry.yaml
```diff
@@ -294,3 +294,33 @@ awp:
...
 		Explosions: piff
 		ValidImpactTypes: Water
+
+NeutronRifle:
+	ValidTargets: Ground
+	ReloadDelay: 120
+	Range: 10c0
+	MinRange: 0c001
+	Report: ichratta.wav
+	Projectile: LaserZap
```
GraionDilach: A TeslaZap DimZap: 1 BrightZap :0 using a custom RGBA palette is closer what RA2 had imo.

GraionDilach:
Will do as a start though. :+1: when the warhead versus fixup is done, changing the weapon type again is optional.

Mailaender:
Updated.

MustaphaTR:
Looks like needs a rebase.

Mailaender:
Rebased.

rules/allied-infantry.yaml
```diff
@@ -382,3 +382,41 @@ jumpjet:
...
+		PipType: Red
+	-Crushable:
+	PortableChrono:
+		ChronoshiftSound: ichrmova.wav
+		DeployCursor: chronosphere
+		DeployBlockedCursor: chronosphere
+		Voice: Move
+	Voiced:
+		VoiceSet: ChronoLegionnaireVoice
+		ChargeDelay: 300
```
abcdefg30: This doesn't look correct.

weapons/allied-infantry.yaml
```diff
@@ -173,3 +173,22 @@ AWPE:
 20mme:
 	Inherits: 20mm
 	Burst: 2
+
+NeutronRifle:
+	ValidTargets: Ground
+	ReloadDelay: 120
+	Range: 10c0
+	MinRange: 0c001
```
abcdefg30: Where does this come from?

Mailaender: https://github.com/OpenRA/ra2/blob/70f14b00d3b3f36e8339255c72d9811610444181/weapons/allied-infantry.yaml#L150 I assume.

GraionDilach: That shouldn't be needed here - jumpjet stuff can crash at range 0 for some reason. CLeg ain't jumpjet.

sequences/allied-infantry.yaml
```diff
@@ -480,3 +479,58 @@ jumpjet:
...
+		Facings: 1
+	die1:
+		Start: 38
+		Length: 15
+		Facings: 8
+	die2:
+		Start: 53
+		Length: 15
+		Facings: 8
+	die3:
```
GraionDilach: These and onwards need an update.

Mailaender:
Updated.

rules/defaults.yaml
```diff
@@ -62,6 +62,17 @@
 	TimedUpgradeBar:
 		Upgrade: invulnerability
 
+^ChronoDisable:
```
MustaphaTR: You should make that effect production, sws, weapons and stuff like that, they would be abit more useful for now. Also the Alpha effect can be added. It must be already possible.

rules/allied-infantry.yaml
```diff
@@ -382,3 +382,41 @@ jumpjet:
...
+	Inherits: ^Infantry
+	Selectable:
+		Bounds: 20, 30, 0, -11
+	Tooltip:
+		Name: Chrono Legionnaire
+		Description: High-tech soldier.\n  Strong vs Infantry, Vehicles\n  Weak vs Aircraft
+	Buildable:
+		Queue: Infantry
+		BuildAtProductionType: Infantry
+		BuildPaletteOrder: 60
```
MustaphaTR: I think this is too low. This must be go to last, as it has TechLevel=10 in original game.

rules/defaults.yaml
GraionDilach: Please read #174.

PR #164, opened 2016-03-13
Added the Guard Border Crossing

Mailaender:
Compare to Tiberian Sun gates https://github.com/OpenRA/OpenRA/pull/9981/ this just comes with one variant.


abcdefg30:
The artwork has a wrong offset, could you fix that?
Otherwise: :+1: / :white_check_mark:

Mailaender:
Added the offset from the tesla coil which is my best guess here.

GraionDilach:
That's a correct offset. :+1:

PR #165, opened 2016-03-13
Fixed CAMERA artwork

Mailaender:
I have no idea why @abcdefg30 added the one from https://github.com/OpenRA/OpenRA/blob/bleed/mods/ra/bits/camera.shp instead of https://github.com/OpenRA/OpenRA/blob/bleed/mods/ts/bits/camera.shp in https://github.com/OpenRA/ra2/pull/120, but I also forgot to check. This fixes the palette and geometry.


abcdefg30:
Ops, thanks. :+1:

PR #169, opened 2016-03-15
Make use of the IsPlayerPalette of NukePowerInfo on namisl

abcdefg30:
Makes use of https://github.com/OpenRA/OpenRA/pull/10923.


GraionDilach:
:+1:

Issue #142, opened 2016-02-18
Tanya and Navy Seals can't swim [$5]


Fincer:
In the original RA2 (Tanya):

![cantswim-ra2-cropped](https://cloud.githubusercontent.com/assets/3705231/13152559/76613eb8-d678-11e5-85e3-dae66cb64a15.gif)

OpenRA/RA2 (Tanya & a group of Navy Seals):

![cantswim-openra2-cropped2](https://cloud.githubusercontent.com/assets/3705231/13152576/8ce26f36-d678-11e5-9ca1-e447d2864b26.gif)

PR #172, opened 2016-03-20
Added swim animations to Tanya and the NAVY Seal

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/142.

### DIFF
--- a/OpenRA.Mods.RA2/OpenRA.Mods.RA2.csproj
+++ b/OpenRA.Mods.RA2/OpenRA.Mods.RA2.csproj
@@ -64,5 +64,6 @@
     <Compile Include="Activities\ChronoResourceTeleport.cs" />
     <Compile Include="Traits\ChronoResourceDelivery.cs" />
     <Compile Include="Traits\Render\WithExitOverlay.cs" />
+    <Compile Include="Traits\Render\WithSwimSuit.cs" />
   </ItemGroup>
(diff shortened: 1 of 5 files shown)


GraionDilach:
This has an issue though, it does not affect `Targetable`, which will mean that units which are only capable of attacking water targets, (Subs, squids, dolphins) will be unable to attack it.

So it might be better if instead of a sequence modifier trait, this will be a "grant-upgrade-if-in-water" trait and the let `WithInfantryBody` and `Targetable` be upgraded with it.

Mailaender:
`WithInfantryBody` is upgradeable, but not `WithDeathAnimation` so I keep doing it with infantry modifiers for now which is simple and robust. I added an upgrade type for underwater targeting though.

GraionDilach:
Nice, since it keeps the Ground targeting in water, that resolves my gameplay issue with it, I'm fine with the infantry sequences visually looking odd then for now.

Notsure if I should wait from a confirmation of @abcdefg30 here with the merging - the :+1: is granted.

Mailaender:
Not sure if we have enough review capacity for a +2 system. I plan to further work on improvements, but that would be OpenRA upstream, not this pull request.

GraionDilach:
On a second thought, I agree. Death animations are a mess anyway. Good job on this regardless.

PR #173, opened 2016-03-22
Added a BountySource badge.

Mailaender:
While @BountySource GitHub integration is a bit lacking https://github.com/bountysource/core/issues/1009, https://github.com/bountysource/core/issues/1010 and https://github.com/bountysource/core/issues/1011 have a badge in the mean time.


GraionDilach:
Should this depend on those aforementioned issues or we should take this as-is now? I'm not certain about it atm.

Mailaender:
It is independent from them or rather filling the gap of currently missing GitHub integration.

abcdefg30:
Yep, looks ok here. :+1:

PR #176, opened 2016-03-27
Balance Update 1.0

vietman391:
Changes:

- Chrono Miner: Decreased speed & capacity, increased HP
- Ore Miner: Ditto, except speed
- Rocketeer: Decreased HP
- Harrier: Decreased HP & Damage (They no longer 1-hit everything)
- Oil Derrick: Decreased capture bonus from 1000 to 500
- Attack Dog: Decreased HP & (possibly) rate of fire (I tested it, nothing really changed)

### DIFF
--- a/rules/aircraft.yaml
+++ b/rules/aircraft.yaml
@@ -1,3 +1,81 @@
+shad:
+	Inherits: ^Aircraft
+	Tooltip:
+		Name: Night Hawk
+		Description: Infantry Transport Helicopter\nUndetectable by radar.\n  Strong vs Infantry\n  Weak vs Vehicles, Aircraft
+	Health:
+		HP: 175
+	Armor:
+		Type: Light
+	Armament:
+		Weapon: BlackHawkCannon
+	Voiced:
+		VoiceSet: NighthawkVoice
+	Buildable:
+		Queue: Vehicle
+		BuildPaletteOrder: 20
+		Prerequisites: ~gaweap
+	Valued:
+		Cost: 1000
+	RevealsShroud:
+		Range: 7c0
+		Type: CenterPosition
+	-AppearsOnRadar:
+	AttackFrontal:
+		Voice: Attack
+	Aircraft:
+		CruiseAltitude: 3072
+		LandWhenIdle: true
+		LandableTerrainTypes: Clear, Rough, Road, DirtRoad
+		TurnSpeed: 10
+		Speed: 100
+		AltitudeVelocity: 120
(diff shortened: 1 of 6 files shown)


GraionDilach:
You see, we are aiming to restore the original values first - which aren't done even. RA2 was the most balanced Westwood C&C actually. Most of your changes are nonsense tbh.

I understand you aim to get what's done playable, but changing these such early is drastic and really makes no sense. I'll have to :-1: this.

BTW, yes, `AttackLeap` also ignores `ReloadDelay`.

vietman391:
It makes more sense when you actually play the mod using these changes, but I understand how small changes like these can make you uncomfortable.

I'm not doing this to give an impression that I'm right about these changes, it's just that I want to see how I can _temporarily_ balance ra2 so that I can get another good stream going while playing ra2 for fun :)

GraionDilach:
The main issue is that this isn't even streamable technically.

Also, you should have realized it by now - since I'll assume you're Spy\` -, that if you end up successfully sneaking in some "temporary" workarounds, after a short while people will forget to revert them and considering how lousy the RA2 mod always was, I'm really not welcoming balance workarounds.

vietman391:
I wasn't the one who started a pull request in the first place; it just so happens Mailaender moved my fork to work for here.

I don't intend for any of my balance changes to pass; I'm just doing this for myself and anyone who's interested.

I understand that you have a personal opinion towards me; but I don't think there is a need for that, and it's not very productive.

Mailaender:
This still seems to base off on an older version of the RA2 mod, so we can't see the real changes nor integrate it. See https://help.github.com/articles/using-git-rebase/ on how to resolve this.

We are indeed only interested in fixes that restore vanilla balancing for now. I assume the harrier one-shot kill every building is a bug for example as the weapon rules don't translate 1:1 into OpenRA.

Phrohdoh:
> We are indeed only interested in fixes that restore vanilla balancing for now.

So is the power-down going to be removed then?

GraionDilach:
Actually the powerdown feature is still less arbitrary than the changes in this PR (for example the miner changes, which are actually an issue of gem worth, which has 250 than the supposed 100 per bail...)

Phrohdoh:
I agree that this shouldn't be taken at this time.
Balance (whatever one may think of it) should definitely match vanilla (as close as sanely possible) before making tweaks.

Mailaender:
We should focus on fixing the rules as in https://github.com/OpenRA/ra2/pull/179 instead changing them. At least for now.

PR #177, opened 2016-03-29
Update to latest bleed

scshunt:
Bleed commit:  728bad95

### DIFF
--- a/rules/player.yaml
+++ b/rules/player.yaml
@@ -3,36 +3,36 @@ Player:
 	TechTree:
 	ClassicProductionQueue@Building:
 		Type: Building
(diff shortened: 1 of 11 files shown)


rules/player.yaml
```diff
@@ -3,36 +3,36 @@ Player:
 	TechTree:
 	ClassicProductionQueue@Building:
 		Type: Building
-		BuildSpeed: .4
+		BuildSpeed: 4000
```
GraionDilach: These should be 40.

Phrohdoh:
:+1: after unstaging the map previews and fixing the `.4` -> `4000` conversion.

PR #178, opened 2016-03-30
Restore the RA2 inis to extra-files.

GraionDilach:
Had enough of jumping back to Phrohdoh/oramod-ra2 all the time.


MustaphaTR:
:+1:

abcdefg30:
Could you elaborate why it is a good idea to store them in the repository?
As they're big files, won't change and you can easily have them locally instead.

Mailaender:
Maybe store them elsewhere if you want to link lines like https://github.com/LipkeGu/TSIniFiles

MustaphaTR:
@Mailaender This has no different than storing them in Phrohdoh/oramod-ra2.

abcdefg30:
I didn't get why we want to store them at all. I (think I) won't object keeping them in here if there's a good reason.

Phrohdoh:
I agree that these files don't need to be stored here, putting them in my repo was just an easy way to refer to them in PRs.

Perhaps wiki pages?

GraionDilach:
I am in the middle of cleaning up the mess this mod has at weapon definitions, and the use of inheritance just makes many of the WW bogus/arbitrary eliteweapon stats more straightforward (for example Flak Trooper having a nerfed AA weapon when reaching elite).

I don't intend to balance these atm but adding comments to the appropriate RA2 line. I really don't want these comments pointing towards an artifact repository.

Wiki pages for these contents are the worst. Even a gist would be more meaningful.

Phrohdoh:
How are wiki pages any worse than a gist?
They are both full git repos and GitHub projects directly correlate to a wiki.

GraionDilach:
Wiki pages are technically documents with Rich Text output compared to the gist being plaintext. INI files should be plaintext.

scshunt:
I'm opposed to this for copyright concerns. We can't put them into a GPL-licensed repository.

GraionDilach:
Meh.

Mailaender:
:bulb: **ProTip:** Run `OpenRA.Utility.exe ra2 --extract rules.ini` to recieve a fresh copy.

PR #179, opened 2016-03-31
Weapon cleanup.

GraionDilach:
Use Inheritance on weapons where possible.
Removed custom Falloff value from Harrier/Black Eagle which caused them delivering 10 times of their supposed damage.
Replaced piff (singular bullet hit) animation with piffpiff (multiple bullet hit) animation where appropriate.
Added a comment to Flak Trooper AA weapon explaining the nerf is caused by Westwood.
Misc ImpactType cleanup on CreateEffectWarheads.
Added missing tesla hit animation to Tesla Trooper and Tesla Coil.
Fixed Crazy Ivan eliteweapon crashing due to an ExplosionPalette typo.

### DIFF
--- a/weapons.yaml
+++ b/weapons.yaml
@@ -14,15 +14,9 @@ UnitExplode:
 		ImpactSounds: expnew09.wav
 
 UnitExplodeSmall:
+	Inherits: UnitExplode
 	Warhead@1Dam: SpreadDamage
-		Spread: 426
 		Damage: 40
-		Versus:
-			None: 90
-			Wood: 75
-			Light: 60
-			Heavy: 25
-		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
 	Warhead@2Eff: CreateEffect
 		Explosions: medium_brnl
 		ImpactSounds: expnew13.wav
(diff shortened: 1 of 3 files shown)


weapons.yaml
```diff
@@ -771,29 +675,13 @@ FlakTrackAAGun:
...
-			None: 150
-			Flak: 80
-			Plate: 50
-			Light: 100
-			Medium: 20
-			Heavy: 20
-			Wood: 0
-			Steel: 0
-			Concrete: 0
 		DamageTypes: BulletDeath
```
abcdefg30: I think you can remove this here too. It'll also inherit the missing "TriggerProne" `DamageTypes` then.

weapons.yaml
```diff
@@ -1054,65 +910,29 @@ HoverMissile:
...
-		DamageTypes: Prone50Percent, TriggerProne, ExplosionDeath
 	Warhead@2Eff: CreateEffect
 		Explosions: medium_grey_explosion
 		ImpactSounds: gexp15a.wav
 		InvalidImpactTypes: Water, Air
-	Warhead@3EffWater: CreateEffect
-		Explosions: small_watersplash
-		ImpactSounds: gexpwasa.wav
-		ValidImpactTypes: Water
 	Warhead@4EffAir: CreateEffect
```
abcdefg30: You can remove this entire warhead when you remove `Air` from `InvalidImpactTypes` of the second warhead.

weapons.yaml
GraionDilach: Now that I see it, it doesn't make a difference because flying infantry should never go prone and this is a strictly-AA weapon.

abcdefg30: Oh, right. My derp. ^^ We might want to remove the "TriggerProne" types then from `FlakTrackAAGun`.

abcdefg30:
Looks good to me otherwise. :+1: / :white_check_mark:

GraionDilach:
Updated.

weapons.yaml
GraionDilach: Done and thereby removed this line as well.

PR #180, opened 2016-04-02
Fix aircraft not showing up on radar

abcdefg30:


GraionDilach:
Whoops. :D :+1:

GraionDilach:
Wait, this needs a fixup, NightHawk was indeed radar-invisible.

abcdefg30:
It already is: https://github.com/OpenRA/ra2/blob/master/rules/aircraft.yaml#L23

PR #185, opened 2016-04-10
Added the Tech Outpost

Mailaender:
without the defensive turret for now.


rules/tech-structures.yaml
```diff
@@ -40,3 +40,49 @@ cahosp:
...
+#	WithRangeCircle:
+#		Range: 12c0
+#		Type: patriot
+#	AutoTarget:
+#	Turreted:
+#		TurnSpeed: 10
+#		InitialFacing: 224
+#		Offset: 550,0,0
+#	AttackTurreted:
+#		Voice: Attack
```
abcdefg30: Um, this actor is not voiced I think.

abcdefg30:
So, what's the problem with repairing and having a turret at the same time?
(I didn't spot something obviously wrong during a quick test.)

Mailaender:
I don't remember. It has been almost a month now. Could be targeting and the rallypoint conflicting. Enabled it nonetheless as it doesn't crash the game or something otherwise serious.

rules/tech-structures.yaml
Mailaender: Removed it.

PR #186, opened 2016-04-10
Added the Tech Airport

Mailaender:
See http://cnc.wikia.com/wiki/Tech_Airport_(Red_Alert_2) for reference.


MustaphaTR:
Works fine but that would be good if you add that to maps that should have.

Mailaender:
Do you know a map (without bridges or other unsupported stuff) that contains one?

MustaphaTR:
Breakaway (which we already have) should have one at middle. I'm not sure about others.

Mailaender:
Updated the Break of Day map. Looks like the civilian buildings from https://github.com/OpenRA/ra2/pull/114 and the lightning rules were missing as well.

abcdefg30:
Looks good to me, except for the prerequisites stuff. Not sure if we can fix that with only yaml changes:
- Having both sides conyards will give you two different paratrooper powers
- (Obvious) Without a conyard you won't get anything

GraionDilach:
Yeah, `SupportPower`s don't have faction tags which needs an upstream change to get that fixed.

abcdefg30:
I think we can take this as is. I have a follow-up pr ready that will take care of the prerequisites issue.

PR #187, opened 2016-04-10
Added tech building notifications

Mailaender:
Followup of https://github.com/OpenRA/ra2/pull/77.


rules/defaults.yaml
```diff
@@ -159,6 +159,10 @@
 		Range: 4c0
 	Tooltip:
 		Description: Tech Building
+	CaptureNotification:
+		Notification: TechBuildingCaptured
+	ActorLostNotification:
```
MustaphaTR: Does this play when actor is destroyed? In RA2 this only plays when anyone else captures our Tech Building iirc.

abcdefg30: This doesn't play at all when the building is captured, but only when it got destroyed.
We might want to expand `CaptureNotification` here to play a sound to the old owner
(if that isn't already possible in some way).

MustaphaTR:
> As I remembered, there's different notifications for different structures in original RA2...

It was in YR. RA2 had all same notifications.

abcdefg30:
I think we should wait for OpenRA/OpenRA#11142 here.

Mailaender:
Superseded by https://github.com/OpenRA/ra2/pull/219.

PR #190, opened 2016-04-18
Added OpenRA.Mods.Common.Traits.Render

Mailaender:
This is required since https://github.com/OpenRA/OpenRA/pull/11122.


abcdefg30:
Thanks.

PR #200, opened 2016-05-05
Night Hawk Added Ore, Gems

jtyle6:
Are able to land on Ore and Gems as in the base game.


GraionDilach:
Ah, right, thankyou.

jtyle6:
I noticed [#11222](https://github.com/OpenRA/OpenRA/pull/11222) Of the Red Alert mod. And did't see in the rules they couldn't land on Ore and gems .

PR #201, opened 2016-05-07
OpenRA.Utility.exe ra2 --upgrade-mod 20160331

Mailaender:
Incorporates the changes from https://github.com/OpenRA/OpenRA/pull/11205 and https://github.com/OpenRA/OpenRA/pull/11144.

Last run was https://github.com/OpenRA/ra2/pull/177.


GraionDilach:
Thanks!

PR #202, opened 2016-05-07
Fixed and enabled the Kirov crash animation

Mailaender:
The out-commented rules had some bit rot in them which I fixed as well.

### DIFF
--- a/rules/aircraft.yaml
+++ b/rules/aircraft.yaml
@@ -121,32 +121,41 @@ zep:
 	AnnounceOnBuild:
 	Voiced:
 		VoiceSet: KirovVoice
-#	SpawnActorOnDeath:
-#		Actor: zephusk
+	SpawnActorOnDeath:
+		Actor: zephusk
 	Targetable@ground:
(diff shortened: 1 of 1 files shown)


rules/aircraft.yaml
```diff
@@ -121,32 +121,41 @@ zep:
...
+	Health:
+		HP: 140
+	Aircraft:
+		TurnSpeed: 10
+		Speed: 50
+		AirborneUpgrades: airborne
+		CanHover: True
+		Voice: Move
+	FallsToEarth:
+		Spins: False
```
abcdefg30: IIRC that is not true, but I had issues to set this up and this probably needs an upstream fix.
(The kirov begins to spin faster and faster which looks very weird.)
Just speaking my "thoughts" out loud.

Mailaender: I left a comment although to me it doesn't make any sense. They aren't helicopters with rear rotors.

abcdefg30:
:+1:

PR #203, opened 2016-05-08
Added Dustbowl from Yuris Revenge

Mailaender:
Closes https://github.com/OpenRA/OpenRA/issues/10686.


abcdefg30:
:+1: Those cows are killing my ears though. :D

jtyle6:
@abcdefg30 yeah there are.

destroy-everything:
I'm pretty sure I killed them all but the noise persists

Mailaender:
If you are sure and can reproduce the error, report it as a bug via https://github.com/OpenRA/ra2/issues/new please.

PR #206, opened 2016-05-15
Polished the Dolphin

Mailaender:
Fixes some oversights from https://github.com/OpenRA/ra2/pull/67.


Mailaender:
Rebased in https://github.com/OpenRA/ra2/pull/220.

Issue #208, opened 2016-05-16
Repository layout doesn't allow releases


Mailaender:
I made some tests and the OpenRA.Mods.RA2.DLLs created cause `TypeLoadExceptions` when used with regular released versions of OpenRA or thirdparty mods.

I assume the hard-coded https://github.com/OpenRA/ra2/tree/master/OpenRA.Mods.RA2/dependencies location is the root of all evil.

It also copies unrelated OpenRA.Game.exe and OpenRA.Mods.*.dll files into the mod root, which isn't wanted behavior.

Also the https://github.com/OpenRA/ra2/wiki/Step-by-Step setup is ridiculously complicated.

GraionDilach:
Either this, either splitting the repository into a ra2-mod and a ra2-dll repository, which can be checked out into the OpenRA and OpenRA/mods folder directly. To fully support such a setup though, one might need to have an additional commit in OpenRA to expand .gitignore to prevent conflicts within the git repos (The same commit can also be used to expand upstream makefiles).

Noidea how well it would ease up producing releases though (tbh until the TS dll is missing from the releases, the discussion around it is IMO moot).

Attacque Supérior and Shattered Paradise already follows this above concept (SP is a better example, because in SP, the mod is indeed in a third git repository, while AS only has the DLL repo set up.) Look at @ABrandau to see how it's done.

Mailaender:
> splitting the repository into a ra2-mod and a ra2-dll repository

Has the downside that both need to be kept in sync and it then needs at least two (three if it also requires an OpenRA upstream change) pull requests even for tiny feature additions.

GraionDilach:
Gitignoring their folders would only need to be done once. Otherwise yes, I know, there isn't a perfect solution.

Mailaender:
Moving the trait logic where it belongs https://github.com/OpenRA/OpenRA/pull/11301 seems to be the only sane way until the libraries can be installed into GAC or distributed via NuGet.

Mailaender:
I followed your suggestions and separated mod logic https://github.com/OpenRA2/Mods.RA2 from the mod ruleset https://github.com/OpenRA2/ra2 The benefit over this repository is that I preserved all commits from previous contributors including:

* https://github.com/Phrohdoh/oramod-ra2
* https://github.com/Phrohdoh/OpenRA/tree/ra2%23master

[...]

PR #209, opened 2016-05-17
Implement the pchote-derived infrastructure.

GraionDilach:
Followup of https://github.com/OpenRA/OpenRA/pull/11301
Depends on https://github.com/OpenRA/OpenRA/pull/10981.
Related to #208.

Makefiles and travis update are still missing and the launchers are placeholders (the shell one was untested even).

For testing purposes, I intended to use https://github.com/GraionDilach/OpenRA/tree/infrastructure-pchote (which is bleed + https://github.com/OpenRA/OpenRA/pull/10981) but as reported in that PR, I got stuck at this point.


GraionDilach:
NTM: Create a revision of https://github.com/Walkman100/OpenRA-Install-Guide/blob/master/Install-RA2/InstallRA2.bat when this gets mergeable.

PR #210, opened 2016-05-19
Update make.ps1

Walkman100:
Allow loading the dependency folder from the command line
This allows running the `make` powershell file completely automated:
```posh
make dependencies "path\to\OpenRA"
```
I'm using this in a [script](https://github.com/Walkman100/OpenRA-Install-Guide/blob/master/Install-RA2/InstallRA2.bat) I have that automatically clones & builds OpenRA and this mod, and currently the user has to [copy & paste](https://github.com/Walkman100/OpenRA-Install-Guide/blob/master/Install-RA2/InstallRA2.bat#L21) the file path that the script outputs into the powershell input.
It would be awesome to be able to do this automatically.

Tested on Win7x64

---
This change is [Reviewable](https://reviewable.io/reviews/openra/ra2/210)

PR #211, opened 2016-05-19
Update make.ps1

Walkman100:
Allow loading the dependency folder from the command line
Followup of #210

---
This change is [Reviewable](https://reviewable.io/reviews/openra/ra2/211)

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -60,7 +60,14 @@ elseif ($command -eq "dependencies")
 {
 	cd OpenRA.Mods.RA2\dependencies
 	$targetDir = $PWD
-	$OpenRADir = Read-Host "Enter the path to your OpenRA install"
+	if ($args.Length -eq 1)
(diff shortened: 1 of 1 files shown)


GraionDilach:
LGTM.

Walkman100:
So, looks like when I tested this I didn't test running it directly - the `if ($args.Length -eq 1)` statement is never run:
```
Test-Path : Cannot bind argument to parameter 'Path' because it is null.
At <path>\ra2\make.ps1:71 char:17
+     if (!(Test-Path <<<<  $OpenRADir))
    + CategoryInfo          : InvalidData: (:) [Test-Path], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorNullNotAllowed,Microsoft.PowerShell.Commands.TestPathCommand
```

PR #212, opened 2016-05-19
Implement the pchote-designed infrastructure.

GraionDilach:
Followup of https://github.com/OpenRA/OpenRA/pull/11301
Depends on https://github.com/OpenRA/OpenRA/pull/10981.
Related to #208.

Makefiles and travis update are still missing and the launchers are placeholders (the shell one was untested even).

For testing purposes, I intended to use https://github.com/GraionDilach/OpenRA/tree/infrastructure-pchote (which is bleed + https://github.com/OpenRA/OpenRA/pull/10981) but as reported in that PR, I got stuck at this point.


launch-RA2.bat
```diff
@@ -0,0 +1 @@
+OpenRA\OpenRA.exe SupportDir=D:/OpenRA-RA2-ipchote/
```
Walkman100: hard-coded `D:/OpenRA-RA2-ipchote/`?

GraionDilach: Placeholder. See the linked upstream issue - atm I left it in as a NTM and as a testcase for upstream to look at.

GraionDilach: I'd gladly have `~dp0` there myself as well but that doesn't work at all atm.

PR #223, opened 2016-05-26
Follow upstream's C4Demolition trait being renamed.

GraionDilach:
Commit II is a freebie - thought I should drop those values afterall.


Mailaender:
Thanks. https://github.com/OpenRA/OpenRA/pull/11294/

PR #228, opened 2016-06-05
Fixed the infantry death animations

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/11391.


rules/defaults.yaml
```diff
@@ -278,13 +278,15 @@
...
 			ExplosionDeath: 1
 			BulletDeath: 2
+			ElectroDeath: 6
 	WithDeathAnimation@effect:
 		DeathSequencePalette: effect
 		DeathPaletteIsPlayerPalette: False
 		DeathTypes:
 			RadiationDeath: 4
-			FlameDeath: 5
-			ElectroDeath: 6
```
GraionDilach: Do not change this, this is valid. electro uses anim.pal and not unittem.pal.

GraionDilach: Ah, we have an SHP override in our bits folder which is indeed converted to unittem - removing that file should be the fix.

GraionDilach: This same applies to flameguy as well.

Mailaender: Removed the SHP workarounds. I didn't know they existed.

rules/misc.yaml
```diff
@@ -105,3 +105,34 @@ CAMERA:
...
+		HP: 160
+	SelfHealing:
+		Step: -10
+		HealIfBelow: 101
+	ScaredyCat:
+	WithDeathAnimation:
+		FallbackSequence: die
+		UseDeathTypeSuffix: false
+		DeathSequencePalette: effect
+		DeathPaletteIsPlayerPalette: False
```
abcdefg30: Minor nit: Please use a lowercase `false` as we/you did in the rest of the file.

abcdefg30:
Looks good to me otherwise. 👍 / ✅
Thanks!

rules/misc.yaml
Mailaender: Fixed.

PR #229, opened 2016-06-05
Replaced UpgradeInWater with the more generic UpgradeOnTerrain.

Mailaender:
See https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Map/TileSet.cs#L65


abcdefg30:
Everything still fine. ✅

PR #230, opened 2016-06-05
Fixed all dependencies being copied in the ra2 mod folder

Mailaender:
This might fix the `TypeLoadExceptions` haunting me in https://github.com/OpenRA/ra2/issues/208.


abcdefg30:
Looks good to me, still compiles and runs fine locally 👍 and travis is also ok. ✅

PR #231, opened 2016-06-13
Removed duplicate bits files

Mailaender:
They exist in sub-directories as well or aren't needed such as the mouse.shp as content installation is now done in the mod chooser.

Another thing that bothers me is the `bits/structures` folder. It contains SHPs, but the reason why they were added there was lost. The commit messages only document the copying of the folder:

* https://github.com/Phrohdoh/OpenRA/tree/ra2%23master/mods/ra2/bits/structures
* https://github.com/Phrohdoh/oramod-ra2/tree/master/ra2/bits/structures


abcdefg30:
Not sure if we want to remove `droads45.tem`. The normal one looks bogus.

abcdefg30:
openra-2016-06-22t205446z

Mailaender:
It is a duplicate of https://github.com/OpenRA/ra2/tree/master/bits/temperate/droads45.tem

abcdefg30:
Oh, interesting. But then the other bits file is invalid and should go as well?

GraionDilach:
Regarding droads45.tem, IIRC that was something i pushed through during my first days because the tileset importer kept crashing on the file.

abcdefg30:
Hm, it didn't crash for me when I removed both files from bits, but I think someone else should confirm I didn't do stupid stuff again.

Mailaender:
Removed it.

abcdefg30:

PR #232, opened 2016-06-13
Added a flat icon with source

Mailaender:
This replaces the current translucent one showing some unclean edges due to improper conversions with a flat one that follows the http://www.openra.net/ style guide. It includes a source file that can be opened wit http://www.getpaint.net/ or https://pinta-project.com/


abcdefg30:
Sorry, I somehow like the old one more.

Issue #124, opened 2016-02-09
Install from original CD [$5]


Mailaender:
Installing from the original CD doesn't work, because Microsoft CAB isn't supported yet.

jtyle6:
@Mailaender do you get? $5 if you close it?

abcdefg30:
@pchote did the actual work. @Mailaender just merged the Pull Request.

Mailaender:
Yep, @pchote please claim https://www.bountysource.com/issues/30699853-install-from-original-cd

PR #233, opened 2016-06-14
Add new content installer metadata to RA2.

pchote:
For use with https://github.com/OpenRA/OpenRA/pull/11375.

Closes https://github.com/OpenRA/ra2/issues/124.

### DIFF
--- a/mod.yaml
+++ b/mod.yaml
@@ -182,13 +182,6 @@ LoadScreen: LogoStripeLoadScreen
 	Image: ra2|uibits/loadscreen.png
 	Text: Carving wooden nickels..., Preparing paratroopers..., Welcoming you to Texas..., Optimizing helium mix..., Receiving thoughts..., Igniting boosters...
 
-ContentInstaller:
-	TestFiles: ^Content/ra2/ra2.mix, ^Content/ra2/language.mix
-	# TODO: Installing from CD doesn't work (because CAB)
-	PackageToExtractFromCD: INSTALL\Game1.CAB
-	ExtractFilesFromCD:
-		.: THEME.mix, ra2.mix, language.mix
-
 ServerTraits:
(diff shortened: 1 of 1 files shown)


Mailaender:
Works. :white_check_mark:

mod.yaml
```diff
@@ -268,3 +261,36 @@ GameSpeeds:
 		Name: Fastest
 		Timestep: 20
 		OrderLatency: 6
+
+ModContent:
+	InstallPromptMessage: Red Alert 2 requires artwork and audio from the original game.
+	HeaderMessage: The original game content must be copied from an original game disc.
+	Packages:
+		base: Base Game Files
+			TestFiles: ^Content/ra2/ra2.mix, ^Content/ra2/language.mix, ^Content/ra2/multi.mix
```
abcdefg30: We don't need (and probably don't want) `mutli.mix`. It just contains `.map` files.

pchote: @Mailaender: did you have a specific reason for including this in the metadata block you posted on https://github.com/OpenRA/OpenRA/pull/11375?

Mailaender: I use `OpenRA.Utility.exe ra2 --extract` etc. regularly which requires these to be mounted.

abcdefg30: > which requires these to be mounted.

Oo why that.

Mailaender: The command uses the virtual file system. That is independent from this though. You can safely remove it for all the other users.

Mailaender:
Merging this now to fix the build.

abcdefg30:
> Merging this now to fix the build.

👎 If you really desperately need to fix the build you could have just fixed @pchote's commit locally and directly push it (closing this pr), IMHO.

Anyway, filed #236.

PR #236, opened 2016-06-18
Remove multi.mix from the install yaml

abcdefg30:
Follow-up of #233.


Mailaender:
That has the downside that developers now need to manually extract this file. An alternative is to use http://ra2maps.zzattack.org/ as a source for .map files.

Issue #245, opened 2016-06-25
Origin support


Mailaender:
Use https://github.com/OpenRA/OpenRA/pull/11522 to grab files from C&C The Ultimate Collection and the free on the house promotion.

PR #246, opened 2016-06-26
Added support for C&C The Ultimate Collection

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/245.

### DIFF
--- a/mod.yaml
+++ b/mod.yaml
@@ -258,12 +258,12 @@ ModContent:
 	Packages:
 		base: Base Game Files
 			TestFiles: ^Content/ra2/ra2.mix, ^Content/ra2/language.mix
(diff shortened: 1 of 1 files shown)


Phrohdoh:
@Mailaender How did you test this?

Mailaender:
With a Windows 7 Origin installation from the 2015 Freeware promotion http://www.pcgamer.com/command-conquer-red-alert-2-is-free-on-origin/

abcdefg30:
Works nicely. Thanks. :) 👍 / ✅

Mailaender:
I tried to set this up as generic as possible. If this doesn't work with the US English version (Origin enforces a localized install) then we need another entry and call this one German.

Issue #239, opened 2016-06-19
PowerShell Error


zerogott:
Hey,
i downloaded the latest master,and the latest openRA bleed,after typing dependencies in the ra2 PowerShell i just get https://i.gyazo.com/5f07a480e473709ddd9ccb6ad71f1dba.png

and info on it?

abcdefg30:
Ah yeah, someone broke the shell within the last weeks. You should be able to bypass the issue by giving the path directly, i.e. just type in "dependencies \<path>" directly. (\<path> is the path to your OpenRA install.)

Walkman100:
https://github.com/OpenRA/ra2/pull/211#issuecomment-221999129

We should either figure out how to fix it (I know nothing about PowerShell scripts) or, if that method is going to be made obsolete by a better repo design anyway, update the wiki page for now.

Mailaender:
As no one came up with a fix I reverted the breaking change https://github.com/OpenRA/ra2/issues/248.

thefloe1:
I guess changing line 63 from
`if ($args.Length -eq 1)`
to
`if ($args.Length -lt 1)`

should do the trick.

Mailaender:
Send a pull request, please.

Phrohdoh:
@Mailaender You already have a PR open that can fix that.

Mailaender:
Closed my pull request. As mentioned above, I just clicked the revert button.

Phrohdoh:
Could a Windows user test #256 please?

Phrohdoh:
Closed via #256.

PR #248, opened 2016-06-27
Fixed PowerShell crashes in make.cmd dependencies

Mailaender:
Reverts OpenRA/ra2#211. Closes https://github.com/OpenRA/ra2/issues/239.


Phrohdoh:
To _really_ fix this you could:
1) Swap the two assignment instructions, or
2) Do what @thefloe1 said [here](https://github.com/OpenRA/ra2/issues/239#issuecomment-228699899)

PR #249, opened 2016-06-28
Fix RadBeamWeapon Color and Width

Phrohdoh:


Mailaender:
Changes look good on first sight, but the zap is invisible for me.

Phrohdoh:
I need to change the ending `00` to `ff` then.

GraionDilach:
Width is calculated in WDist since the linerenderable update, so I guess 2 was too small. I personally tend to use a value of 25 in AS.

Issue #222, opened 2016-05-26
Crash when creating power plant


davidacampos:
Not sure if this is an ra2 or OpenRA issue. Whenever I start a game, the first thing after deploying the base is building a power plant. When it finishes building the plant it throws the following exception:

```
Red Alert 2 Mod at Version {DEV_VERSION}
on map dae1d2a018fde88fae2cbf502823ff0918a132d4 (Breakaway by Westwood Studios).
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.42000
Exception of type "System.InvalidOperationException": Actor does not have trait of type "OpenRA.Mods.Common.Traits.PowerManager"
   at OpenRA.TraitDictionary.TraitContainer"1.Get(UInt32 actor) in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\TraitDictionary.cs:line 151
   at OpenRA.TraitDictionary.Get[T](Actor actor) in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\TraitDictionary.cs:line 89
   at OpenRA.Actor.Trait[T]() in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\Actor.cs:line 247
   at OpenRA.Mods.RA.Traits.InfiltrateForPowerOutage..ctor(Actor self, InfiltrateForPowerOutageInfo info)
   at OpenRA.Mods.RA.Traits.InfiltrateForPowerOutageInfo.Create(ActorInitializer init)
   at OpenRA.Actor..ctor(World world, String name, TypeDictionary initDict) in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\Actor.cs:line 102
   at OpenRA.World.CreateActor(Boolean addToWorld, String name, TypeDictionary initDict) in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\World.cs:line 247
   at OpenRA.World.CreateActor(String name, TypeDictionary initDict) in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\World.cs:line 242
   at OpenRA.Mods.Common.Traits.PlaceBuilding.<>c__DisplayClass6.<ResolveOrder>b__2(World w)
   at OpenRA.World.Tick() in c:\Users\<user>\Downloads\OpenRA-bleed\OpenRA.Game\World.cs:line 326
[...]

PR #253, opened 2016-06-29
Add fetch-deps script

Phrohdoh:
Hopefully this can ease the "are you running against the _latest_ OpenRA" question.
ref https://github.com/OpenRA/ra2/issues/222#issuecomment-229373017


abcdefg30:
#267 was merged instead.

Issue #240, opened 2016-06-19
Civilian building husks


Mailaender:
All civilian buildings leave flat crumble found in their sprite that is rendered with the terrain palette.

PR #254, opened 2016-06-29
Added civilian building husks

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/240.

### DIFF
--- a/rules/civilian-props.yaml
+++ b/rules/civilian-props.yaml
@@ -413,20 +413,6 @@ spkr01:
 	Tooltip:
 		Name: Drive-In Speaker
 
-carus02a:
-	Inherits: ^CivBuilding
-	Armor:
-		Type: steel
-	Health:
-		HP: 800
-
-carus02b:
-	Inherits: ^CivBuilding
-	Armor:
-		Type: steel
-	Health:
-		HP: 500
-
 carus02c:
 	Inherits: ^CivBuilding
 	Armor:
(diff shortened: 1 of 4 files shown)


MustaphaTR:
Should we call them as Husk here? They are called as Rubble in original game.

Code looks fine but we need to check if you added rubble for something rubbleless first. Some buildings doesn't have one.

Phrohdoh:
Does `^CivBuildingHusk` need `AutoSelectionSize` (inherited from `^SpriteActor`)?

I'm behind the idea of renaming husk to rubble here (and then dead->rubble in the sequences) but that isn't absolutely necessary.

Could actors walk over rubble in vanilla RA2? (I've not had any luck running the game via Wine)
Other than those little nits, this is very cool. Looks good ingame.

screen shot 2016-06-30 at 12 25 45 pm

MustaphaTR:
Units can walk over Rubbles in vanilla RA2. Rubbles also shouldn't have a selection box.

Phrohdoh:
They aren't selectable with this PR.

Mailaender:
`AutoSelectionSize` is required for the rendering bounds only. The trait name is a bit misleading. It has nothing to do with selection boxes.

I renamed everything to rubble and allowed units to walk over it.

abcdefg30:
`Exception of type 'System.InvalidOperationException': Unit 'carus02g' does not have a sequence named 'rubble'`

abcdefg30:
Looks quite fine otherwise, but not sure how many more actors are affected by the crash.

Mailaender:
Fixed. The problem occurs when TEMPERAT and SNOW sequences don't match.

abcdefg30:
Looks good to me now. 👍

PR #255, opened 2016-06-30
Fixed laser zap beam width

Mailaender:
Looks like the upgrade rule from https://github.com/OpenRA/OpenRA/pull/10148/ never ran here.

### DIFF
--- a/weapons.yaml
+++ b/weapons.yaml
@@ -990,7 +990,7 @@ Comet:
 	Range: 10c0
 	Report: vpriatta.wav
 	Projectile: LaserZap
-		BeamWidth: 5
+		Width: 50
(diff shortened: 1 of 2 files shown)


GraionDilach:
The contrails weren't either, if you're fixing linerenderables.

Phrohdoh:
Could you fix this as well please?
```
weapons/soviet-infantry.yaml
36:     Projectile: LaserZap
```

Mailaender:
> Could you fix this as well please?

Didn't I already do that?

Phrohdoh:
Indeed, sorry. I don't recall seeing that previously.

abcdefg30:
Thanks.

PR #256, opened 2016-06-30
Ask for input if 'dependencies' has no args passed [make.ps1]

Phrohdoh:
ref #239


abcdefg30:
Thanks. Should have fixed that quite some time ago myself!

Issue #174, opened 2016-03-23
Chrono Legionairre - tasklist


GraionDilach:
This is a tasklist/tracking issue for Chrono Legionairre implementation. Mostly for me, I guess but also better to be written, because it'll require some rewirings within the mod to make it work as it should.

- [ ] Add the Chrono Legionnaire into the sequences/enable it at the gameplay #162 
- [ ] Add the Chrono Legionnaire jump
  - The trait should check for `IsMoving` when it's yes, calculate the distance between current position and destination, move the actor to destination, then divide the distance with a constant defined on the trait and apply upgrades to the actor with the result used as duration.
    - Incase of the CLeg, the best would be an EMP-alike (`RejectsOrders` even?) upgrade kit on it.
- [ ] Revamp `Targetable` upgrades from `aircraft`/`airborne`/`parachute` to `disableground` and `enableair`. Allow all ground actors to be upgradable with `disableground`.
    - This can simplify parachuting actors as well, since technically in RA2 all ground infantry and units were parachutable (with the parachute granting `disableground`, `enableair` and `parachute`).
- [ ] Add a new `DeathType` to all weapons, say `DefaultDeath`. Change explosions to rely on crushing/C4/`DefaultDeath` killer weapon.
    - This might even need a patch upstream due to `WithBuildingExplosion`.
- [ ] Add the CLeg weapon.
  - Add a `Targetable@temporal` trait to all (ground?) actors
  - Let the CLeg have 2 weapons, one which grants `disableground` and `temporal` to all ground targets (while affecting `Ground` and `Temporal`), and the weapon proposed in #162 should only affect `Temporal` targets.
    - In RA2, the temporal effect disabled all functions, including production and whatnot. Might be easier to transfer the target ownership to hostile neutral (Creeps?) to achieve this as part of the first weapon than to say, add upgrade support to `Production(Queue)`. The temporal upgrade otherwise can be reused from the jumping trait from the second step.
    - The CLeg weapons should not have `DefaultDeath` at all, but strictly a `TemporalDeath`, preventing death explosions to happen - as it was in RA2 - and allowing the `warpaway` animation to be played on all actors, through say an `Explosion@temporal` trait which has `TemporalDeath` as `DeathType`.
- [ ] Add the sinus beam rendering.

sorcerer86pt:
How the status for the other tasks?

MustaphaTR:
Sinus beam is already in: #656

Mailaender:
[...]

PR #258, opened 2016-07-01
Add the Chrono Legionnaire

abcdefg30:
A revived version of #162.
First step towards finishing #174.

### DIFF
--- a/cursors.yaml
+++ b/cursors.yaml
@@ -195,6 +195,9 @@ Cursors:
 		chronosphere:
 			Start: 357
 			Length: 12
+		chrono-target:
+			Start: 357
+			Length: 12
 		remove-bomb:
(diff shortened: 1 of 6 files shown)


MustaphaTR:
As i said before in old PR, i think making some other things disabled too should be possible. This PR only have mobile tag disabled. Units/buildings being erased can't fire, provide SW, provide power, build units. But can provide prerequisite iirc. Also anims freeze during this effect. Some of these should be possible to add in this PR.

Phrohdoh:
Does this include the continuous-fire and temporal logic?

Mailaender:
Obviously not. This is just the sequence setup. It doesn't add any logic. Weaponry is a placeholder.

PR #260, opened 2016-07-02
Added asset browser support for the Urban tileset

Mailaender:
Sequences like https://github.com/OpenRA/ra2/pull/259 are otherwise impossible to spot.


abcdefg30:
Thanks. Nice cleanup.

PR #262, opened 2016-07-02
Add more exits to gayard and nayard

### LINKED ISSUE #261: Vessel producers need more exits
It is easy to accidentally penalize yourself for something you shouldn't be penalized for.
ref #250.

screen shot 2016-07-02 at 2 44 51 pm

screen shot 2016-07-02 at 3 00 37 pm

Phrohdoh:
Closes #261

screen shot 2016-07-02 at 4 12 37 pm

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -424,14 +424,74 @@ gayard:
 		Type: Concrete
 	RevealsShroud:
 		Range: 10c0
-	Exit@1:
-		SpawnOffset: 0,-113,0
-		Facing: 96
-		ExitCell: 5,4
-	Exit@2:
+	# X -2 column
+	Exit@n2_n2:
+		ExitCell: -2,-2
+	Exit@n2_n1:
+		ExitCell: -2,-1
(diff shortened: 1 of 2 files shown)


Mailaender:
It might look a bit more polished if you restrict it to NE, SE, SW and NW so the boats don't collide with the supporting pillars.

Phrohdoh:
Are you also suggestion I remove the spawn offsets then? Those offsets will cause exactly that.

Mailaender:
I suggest to set exits at `-2, 0`, `0,2` `4,1` and `1,4` instead.

Mailaender:
Or wait now I understand it. You need the large exits to check for land masses. Better set the radius wider on the northern part as well if someone places it with beach head above it.

Phrohdoh:
I don't understand.
The radius is the same all the way around, 2.

I don't see any reason it couldn't be 1 but 2 was already established so I went with that.

Phrohdoh:
screen shot 2016-07-04 at 8 17 29 am

abcdefg30:

PR #265, opened 2016-07-03
Fixed starting units showing up as "Unlabeled"

Mailaender:
Looks like this weird typo now causes trouble.


abcdefg30:
👍 / ✅

MustaphaTR:
Thanks for fixing. Possbly i did the typo. :+1:

PR #267, opened 2016-07-03
Implement Cake-based build system

Phrohdoh:
This introduces a dependency on [`cake`](http://cakebuild.net/) for those of us who want an easy compilation via `$ make`.
***It does not affect anyone else!***

Compiling via an IDE (or directly with msbuild/xbuild) is unchanged.

I think this is a fair trade-off, especially given the fact not using it doesn't make compilation any more complex than it already is, and doesn't change anything for those who don't know anything about it.

Windows users will want to grab [the `0.13.0` release](https://github.com/cake-build/cake/releases/download/v0.13.0/Cake-bin-v0.13.0.zip) or `choco install cake.portable`.
OS X brew users can `brew install cake`.
Linux people are on their own (just grab the above release and throw it on your `$PATH`?)

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -1,146 +1,26 @@
-function FindMSBuild
-{
-	$msBuildVersions = @("4.0")
-	foreach ($msBuildVersion in $msBuildVersions)
-	{
-		$key = "HKLM:\SOFTWARE\Microsoft\MSBuild\ToolsVersions\{0}" -f $msBuildVersion
-		$property = Get-ItemProperty $key -ErrorAction SilentlyContinue
-		if ($property -eq $null -or $property.MSBuildToolsPath -eq $null)
-		{
-			continue
-		}
-		$path = Join-Path $property.MSBuildToolsPath -ChildPath "MSBuild.exe"
-		if (Test-Path $path)
-		{
-			return $path
-		}
-	}
-	return $null
-}
-
 if ($args.Length -eq 0)
 {
-	echo "Command list:"
-	echo ""
-	echo "  all             Builds the mod dll."
-	echo "  dependencies    Copies the mod's dependencies into the"
-	echo "                  'OpenRA.Mods.RA2/dependencies' folder."
-	echo ""
-	$command = (Read-Host "Enter command").Split(' ', 2)
+	$command = "default"
 }
-else
+elseif ($args.Length -eq 1)
 {
-	$command = $args
+    $command = $args
+
+    if ($command -eq "help"   -or
+        $command -eq "h"      -or
+        $command -eq "--help" -or
+        $command -eq "-h"     -or
+        $command -eq "/h"     -or
+        $command -eq "/?")
+    {
+        echo "Target list:"
+        echo ""
+        echo "  default   Builds the mod dll."
+        echo "  deps      Copies the mod's dependencies into the"
+        echo "              'OpenRA.Mods.RA2/dependencies' folder."
+        echo ""
+        exit
+    }
 }
 
-if ($command -eq "all")
-{
-	$msBuild = FindMSBuild
-	$msBuildArguments = "/t:Rebuild /nr:false"
-	if ($msBuild -eq $null)
-	{
-		echo "Unable to locate an appropriate version of MSBuild."
-	}
-	else
-	{
-		cd OpenRA.Mods.RA2
-		$proc = Start-Process $msBuild $msBuildArguments -NoNewWindow -PassThru -Wait
-		if ($proc.ExitCode -ne 0)
-		{
-			echo "Build failed. If just the development tools failed to build, try installing Visual Studio. You may also still be able to run the game."
-		}
-		else
-		{
-			echo "Build succeeded."
-		}
-		cd ..
-	}
-}
-elseif ($command -eq "dependencies")
-{
-	cd OpenRA.Mods.RA2\dependencies
-	$targetDir = $PWD
-	if ($args.Length -eq 0)
-	{
-		$OpenRADir = Read-Host "Enter the path to your OpenRA install"
-	}
-	else
-	{
-		$OpenRADir = $command[1]
-	}
-	if (!(Test-Path $OpenRADir))
-	{
-		echo "Given directory does not exist!"
-	}
-	else
-	{
-		cd $OpenRADir
-
-		if (!(Test-Path OpenRA.Game.exe))
-		{
-			echo "Unable to find 'OpenRA.Game.exe'!"
-			echo "You need to build OpenRA first."
-		}
-		else
-		{
-			cp OpenRA.Game.exe $targetDir
-		}
-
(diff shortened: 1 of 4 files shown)


make.ps1
```diff
@@ -1,146 +1,21 @@
...
-	$command = (Read-Host "Enter command").Split(' ', 2)
-}
-else
-{
-	$command = $args
+	$command = "default"
 }
-
-if ($command -eq "all")
+else if ($args.Length -eq 1)
```
abcdefg30: `else if` -> `elseif`.

make.ps1
```diff
@@ -1,146 +1,21 @@
...
-		}
-		else
-		{
-			echo "Build succeeded."
-		}
-		cd ..
-	}
+    $command = $args
+
+    if ($command == "help" -or $command == "h" -or $command == "--help" -or $command == "-h" -or $command == "/h" -or $command == "/?")
```
abcdefg30: You need to replace all `==` by `-eq`.

make.ps1
```diff
@@ -1,146 +1,21 @@
...
+    {
+        echo "Target list:"
+        echo ""
+        echo "  default   Builds the mod dll."
+        echo "  deps      Copies the mod's dependencies into the"
+        echo "              'OpenRA.Mods.RA2/dependencies' folder."
+        echo ""
+	exit
```
abcdefg30: Wrong indentation.
I'd also suggest you add a `pause` statement in a seperate line above the `exit`.

make.ps1
```diff
@@ -1,146 +1,21 @@
...
-	while ($true)
-	{
-		if ([System.Console]::KeyAvailable)
-		{
-			break
-		}
-		Start-Sleep -Milliseconds 50
-	}
-}
+cake $command
```
abcdefg30: I could only get this to run without specifying the full path by using `start build.cake -target="$command"`.
(Which will by default open an "open with" dialog, where the user has to select `cake.exe` then.)

abcdefg30:
Somehow your `$"..."` notation doesn't work for me:
`Error: build.cake(79,19): error CS1056: Unexpected character '$'`

make.ps1
Phrohdoh: Why pause then exit?
From a user standpoint all that does is require another keystroke before I can start the build.

abcdefg30: You can't see the help page otherwise (as it immediately closes).

make.ps1
Phrohdoh: Is `cake` on your `%PATH%`?

make.ps1
Phrohdoh: This closes the host process, not just the script?

make.ps1
abcdefg30: I would have to add that manually to `%PATH%`.

make.ps1
abcdefg30: Actually, you're right. It shouldn't matter too much as you usually call that using a normal consol window (kinda a separate process) anyway.

make.ps1
Phrohdoh: Right, this is true of every OS.

abcdefg30: Any way: This needs to be `-target="$command` instead of just `$command`.

Phrohdoh:
Updated.

abcdefg30:
```
An error occured when executing task 'deps'.
Error: A part of the path "OpenRA.Mods.RA2/dependencies\Eluant.dll" could not be found.
```
After it asked me wether I want to copy the `Eluant.dll` over.

abcdefg30:
Never mind, user error as I had no `dependencies` folder.

abcdefg30:
I'd like for at least one other 👍 (or simple "I'm ok with it") from @Mailaender and/or @GraionDilach.

Mailaender:
I can't test this right now, but I support the choice :cake: which looks like a sane solution. :+1:

GraionDilach:
I don't like the idea of introducing a dependency upstream lacks to Windows users for compilation.

Phrohdoh:
You aren't required to use cake, msbuild will work just fine.
This isn't a Windows-only solution, Cake works just fine on OS X and Linux.

GraionDilach:
You are required to use cake if you don't have Visual Studio because you're explicitly retangling make.ps1 to use it. Cake is not a part of Microsoft Windows nor upstream OpenRA.

You're introducing a dependency for Windows users just for the sake of it.

Phrohdoh:
MSBuild is shipped with the .NET Framework, every Windows user will have it.

abcdefg30:
> MSBuild is shipped with the .NET Framework, every Windows user will have it.

True, in theory every windows user could simple run one line in PowerShell. But it's not as convenient as having a script (and not having to have the knowledge yourself). This kinda makes Cake compulsory for the "standard" user then, I guess.

samrocketman:
Why not try to detect cake in `make.ps1` and if cake doesn't exist fall bake to MSBuild?  That seems like a sane behavior.

samrocketman:
Also, on all platforms `Makefile` is typically capitalized.

samrocketman:
> You aren't required to use cake, msbuild will work just fine.

In the README you put:

> You must have `cake`...

You're not really disclosing that MSBuild will work just fine to potential devs.  I think the MSBuild command should at least be documented.

Phrohdoh:
> You must have cake installed if you want to build this via the makefiles provided (make.ps1, makefile).

Of course MSBuild will work.. that is how .NET projects / solutions are compiled.
I don't think this project (or every other .net/mono project) needs to reiterate that.

samrocketman:
I tend to think from the perspective of someone who's new to the technology.  Why shouldn't every .NET or mono project iterate over the ways it should be built?

For me, when I encounter a project; how well documentation is written makes or breaks whether or not I'll contribute.  I'm not saying that about this project but in general.  In any case, just wanted to throw out my thoughts for consideration.  Mainly about detecting cake and falling back.

Phrohdoh:
I see no reason not to do that, unfortunately I do not have the time to implement myself.

Phrohdoh:
This is a bit tricky as the `build.cake` does more than invoking MSBuild.
With input from the user it also copies dependencies.
Cake is also used on *nix here and has the mind to invoke XBuild if running on non-Windows.

The best approach (if we are going to keep this Cake system, perhaps it wasn't a great idea for me to implement it in the first place) would probably be to determine if `cake` is installed in `make.ps1`.

If it is go forward as usual.

Otherwise error out saying something like:

```
Cake isn't installed! You can proceed to build the RA2 mod by either:
    * Installing Cake from <url> then rerunning this file
    * Build via Visual Studio / SharpDevelop / MonoDevelop / etc
    * Run MSBuild <link to msbuild docs / instructions?>
```

Maybe make those actionable by assigning them numbers and taking input from the user.

If either of the first two are chosen just exit (the user has manual work to do then).

If the last is chosen then find MSBuild as we were previously and invoke it for them.

Issue #181, opened 2016-04-02
The word "primary" is drawn in white on the white terrain and consequently is almost invisible.


scshunt:
See OpenRA/OpenRA#10996. The bug *does* exist in RA2 :)

Mailaender:
The problem is that we are using RA95 artwork as placeholders instead of engine generated texts that are better readable.

Phrohdoh:
So which way do you want to go, rendered text or artwork?

GraionDilach:
I plan to port https://github.com/OpenRA/OpenRA/pull/11424 over but right now I am still suffering from the "terrain-sprite-sheet-too-big-for-Intel-HD-2000" crash on map load with still no idea when I can resolve that issue on hardware side.

Mailaender:
A nicer fix would be a new configurable text rendering trait to get rid of the placeholder SHPs.

Phrohdoh:
I just threw this together.
We could make a `WithTextDecoration` trait.

```diff
diff --git a/OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs b/OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs
index ba9c11a..6c25d5f 100644
--- a/OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs
+++ b/OpenRA.Mods.Common/Traits/Buildings/PrimaryBuilding.cs
@@ -10,7 +10,10 @@
 #endregion
 
 using System.Collections.Generic;
+using System.Drawing;
 using System.Linq;
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Graphics;
 using OpenRA.Mods.Common.Orders;
 using OpenRA.Traits;
 
@@ -34,20 +37,54 @@ public class PrimaryBuildingInfo : ITraitInfo, Requires<UpgradeManagerInfo>
 		[Desc("The speech notification to play when selecting a primary building.")]
 		public readonly string SelectionNotification = "PrimaryBuildingSelected";
 
+		public readonly bool ShouldRenderText = false;
+		public readonly string Text = "Primary";
+		public readonly bool RequireSelectionForText = true;
+		public readonly string Font = "Small";
+		public readonly bool UsePlayerColor = false;
+
+		[FieldLoader.LoadUsing("LoadTextColors")]
+		public readonly Dictionary<string, Color> TextColors;
+
+		static object LoadTextColors(MiniYaml y)
+		{
[...]

PR #271, opened 2016-07-04
Fixed the primary building rendering

Mailaender:
Basically https://github.com/OpenRA/OpenRA/pull/11565/ adapted for RA2. Fixes https://github.com/OpenRA/ra2/issues/181.

### DIFF
--- a/rules/allied-structures.yaml
+++ b/rules/allied-structures.yaml
@@ -137,10 +137,9 @@ gapile:
 	Production:
 		Produces: Infantry
 	PrimaryBuilding:
-	WithDecoration@primary:
+	WithTextDecoration@primary:
 		RequiresSelection: true
-		Image: pips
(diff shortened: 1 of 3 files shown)


MustaphaTR:
I didn't test but i think this renders them like in TS and this is not the true way to go because RA2 primary/power text is different. Gimme a sec to find an image.

MustaphaTR:
This is power one but it is similar: https://github.com/OpenRA/OpenRA/issues/6011

Phrohdoh:
I agree that implementing the banner would be nice (possibly an addition to `WithTextDecoration`) but this looks good enough, so 👍.

screen shot 2016-07-05 at 8 37 26 am

MustaphaTR:
This looks fine for now so :+1: from me too. We may need to open an issue to discuss about the banner.

Issue #270, opened 2016-07-04
cathosp is missing


Mailaender:
It looks like the original game is differentiating between civilian hospital and tech hospital: https://github.com/Phrohdoh/oramod-ra2/blob/master/extra-files/rules.ini#L10548

MustaphaTR:
Do we need it. I don't think so.  I think only its name is different(i can't check the rules now,  my browser has problems with big files) and some maps uses so its unexistance cause problems at importing. Nothing else.

GraionDilach:
Merge them in the importer. One of these were unused.

Mailaender:
https://www.diffchecker.com/9cbnazh8 they are indeed similar. Explosion effects seems to be the only difference. I thought one was decorational and the other functional.

Phrohdoh:
It is possible that someone wrote custom rules in their maps for the unused hospital type.
By merging them without care you'd be punishing them.

This may not be an issue in practice but it should be considered.

PR #272, opened 2016-07-05
Added the Tech Hospital

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/270#issuecomment-230362189.

### DIFF
--- a/rules/tech-structures.yaml
+++ b/rules/tech-structures.yaml
@@ -85,3 +85,10 @@ cahosp:
 		StartRepairingNotification:
 	RallyPoint:
 	WithIdleOverlay:
+
+cathosp:
+	Inherits: cahosp
+	EditorOnlyTooltip:
+		Name: Tech Hospital
+	RenderSprites:
+		Image: cahosp


MustaphaTR:
This is unnecessary. You remember the compression link u gave. UIName= is whar matters not Name=. 

If you are not gonna edit rubble/debris don't add it at all.

Phrohdoh:
This is not unnecessary.
If someone has used this actor we should support that.

MustaphaTR:
Maybe noy unnecessary,  but wrong. As i said they both have same ingame name.

Phrohdoh:
@Mailaender Could you remove `Tooltip` from this so we can get this merged?

Mailaender:
I believe it makes sense to differentiate via the tooltip in both the map editor and in-game.

Phrohdoh:
Is there really any difference though?
They seem to serve the same purpose (if the new one serves any purpose at all).

The tooltip in the map editor will show `cathosp` so it differentiates already.

MustaphaTR:
Agreed with @phrohdoh . Codename is already shown pn editor.

Mailaender:
I changed it to `EditorOnlyTooltip` now.

Phrohdoh:
That in no way addresses applying a new name that the original didn't have to the actor.

abcdefg30:
I think it's a good idea. Mappers might wonder why the hospital is "duplicated". They would need to look the differences in the rules between those two up anyways (to see if there any, and if yes which), both with and without a tooltip.

abcdefg30:
Hm, I just discovered that there is a `cathosp.shp`. Maybe we can use that (although the original didn't)?

GraionDilach:
cathosp.shp was never used in the actual game, being an unfinished asset (lacking damage frame, has a mess at palettes, with citytem.pal being used for snow and citysno.pal used for temperate IIRC, etc).

abcdefg30:
Ah well, seems like I accidently had the wrong palette selected when discovering the shp. Thanks anyway.

So, this looks fine to me now. 👍 / ✅

PR #276, opened 2016-07-11
Clean the death sequence setup up

abcdefg30:
- Removed the redundant `die5` sequence
- Remove die-crushed as we don't use* it and the original doesn't have it

*[search?q=CrushedSequence](https://github.com/OpenRA/ra2/search?utf8=%E2%9C%93&q=CrushedSequence) shows no results and [the default is `null`](https://github.com/OpenRA/OpenRA/blob/d97d02b97f2855bddd17d51bc3b917bfdcee21c2/OpenRA.Mods.Common/Traits/Render/WithDeathAnimation.cs#L36).

### DIFF
--- a/sequences/allied-infantry.yaml
+++ b/sequences/allied-infantry.yaml
@@ -78,9 +78,6 @@ adog:
 	die4: # todo: copy of die1
 		Start: 87
 		Length: 14
-	die5: # todo: copy of die1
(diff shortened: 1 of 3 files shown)


Phrohdoh:
👍

GraionDilach:
And another :+1:.

PR #278, opened 2016-07-14
Fix the tech-building flags of all tech-structures

abcdefg30:
caoild
caairp
cahosp


Mailaender:
I guess you could give the flags a slightly higher Z offset: 
image

abcdefg30:
Rebased and updated.

abcdefg30:
caoildz

PR #283, opened 2016-07-25
Add 'clean' and 'dependencies' target to makefiles

Phrohdoh:


makefile
```diff
@@ -1,5 +1,11 @@
 default:
-	cake -target=default
+		cake -target=default
 
 deps:
-	cake -target=deps
+		cake -target=deps
+
+dependencies:
```
chrisforbes: You should really have a `.PHONY` rule that lists these targets, so make doesn't think these rules produce a file of the target name.

Phrohdoh: Ok, I'll add that to this PR later.

Phrohdoh:
I've added the `.PHONY` rule.

abcdefg30:
You didn't update `build.cake` (`Error: The target 'clean' was not found.`)?

Phrohdoh:
Updated again. I forgot to actually include `clean` in this script.

abcdefg30:
Just for the record: We still want to also clean the dependencies.

Phrohdoh:
This now removes `OpenRA.Mods.RA2/dependencies/*.(exe|dll)`.

abcdefg30:
You need to add the redirect for `dependencies` -> `deps` in windows. At the moment I get "Error: The target 'dependencies' was not found.".

Phrohdoh:
I am really off of my game as of late.

Phrohdoh:
I've updated this:

* Fixed a syntax error in make.ps1

abcdefg30:
Looks good to me now. :+1: / :white_check_mark: Thanks.

PR #284, opened 2016-07-25
Remove ingame-infostats and ingame-observerstats from the chrome definitons

abcdefg30:
Those seemed to be copied from ra anyway.
This takes care of OpenRA/OpenRA#11645 and OpenRA/OpenRA#11627.

openra-2016-07-25t203940z
openra-2016-07-25t203945z
openra-2016-07-25t203950z


Mailaender:
Thanks for the cleanup.

Issue #287, opened 2016-08-05
No Game content after running fetch-content.sh


yooouuri:
After running fetch-content.sh, files got downloaded to ~/.openra/Content/ra2/

```
MacBook-Pro-van-Youri:ra2 youri$ bash fetch-content.sh 
Downloading RA2 mod content
~/.openra/Content/ra2 ~/Desktop/OpenRA/mods/ra2
--2016-08-06 01:54:22--  http://xwis.net/downloads/Red-Alert-2-Multiplayer.exe
Resolving xwis.net... 46.105.123.98
Connecting to xwis.net|46.105.123.98|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 206530229 (197M) [application/x-msdos-program]
Saving to: ‘Red-Alert-2-Multiplayer.exe’

Red-Alert-2-Multipl 100%[===================>] 196.96M  11.2MB/s    in 18s     

2016-08-06 01:54:40 (10.9 MB/s) - ‘Red-Alert-2-Multiplayer.exe’ saved [206530229/206530229]


7-Zip [64] 15.14 : Copyright (c) 1999-2015 Igor Pavlov : 2015-12-31
p7zip Version 15.14.1 (locale=utf8,Utf16=on,HugeFiles=on,64 bits,8 CPUs x64)

Scanning the drive for archives:
1 file, 206530229 bytes (197 MiB)

Extracting archive: Red-Alert-2-Multiplayer.exe
--       
Path = Red-Alert-2-Multiplayer.exe
Type = Nsis
Physical Size = 206530229
Method = LZMA:23
Solid = +
Headers Size = 10830
Embedded Stub Size = 33280
SubType = NSIS-2

Everything is Ok          

Files: 179
Size:       472948959
Compressed: 206530229
~/Desktop/OpenRA/mods/ra2
MacBook-Pro-van-Youri:ra2 youri$
```

When running openRA it still gives me an error with "Original Game Content must be copied..."

Mailaender:
Can you post the console output from:
```shell
cd ~/.openra/Content/ra2/
ls
```

yooouuri:
```
MacBook-Pro-van-Youri:ra2 youri$ ls -all
total 854960
drwxr-xr-x  8 youri  staff        272  6 aug 01:54 .
drwxr-xr-x  3 youri  staff        102  6 aug 01:54 ..
-rw-r--r--  1 youri  staff   53116040 22 mei  2001 language.mix
[...]

PR #288, opened 2016-08-06
Fetch-content: Check for OS and change the download directory.

yooouuri:
Fixes https://github.com/OpenRA/ra2/issues/287.

### DIFF
--- a/fetch-content.sh
+++ b/fetch-content.sh
@@ -3,10 +3,18 @@
 # Die on any error for Travis CI to automatically retry:
 set -e
 
-if [ ! -e ~/.openra/Content/ra2/ra2.mix ]; then
+if [ "$(uname)" == "Darwin" ]; then
+	DIR="$HOME/Library/Application Support/OpenRA/Content/ra2/"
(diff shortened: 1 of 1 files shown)


fetch-content.sh
```diff
@@ -3,10 +3,18 @@
 # Die on any error for Travis CI to automatically retry:
 set -e
 
-if [ ! -e ~/.openra/Content/ra2/ra2.mix ]; then
+if [ "$(uname)" == "Darwin" ]; then
+	DIR="$HOME/Library/Application Support/OpenRA/Content/ra2/"
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
```
Walkman100: maybe it would be simpler to put `else` here, because if it's not on Mac OS X then it's on Linux.

Phrohdoh:
@abcdefg30 Do you want to merge this as-is and change the elif to else in a followup commit or should I open a PR  with the requested change?

abcdefg30:
Added in https://github.com/OpenRA/ra2/commit/ad3075002934ef0ae3e31e8c6f5f95b90ecf5eb8.

PR #291, opened 2016-09-03
Updated ChromeMetrics ContrastColor definitions

Mailaender:
Fixes https://github.com/OpenRA/OpenRA/pull/11693#issuecomment-244544476.


abcdefg30:
Sorry for the delay. https://github.com/OpenRA/OpenRA/pull/11975 isn't going anywhere soon as it seems. :(

PR #293, opened 2016-09-05
Update metrics file for OpenRA bleed

alercah:
As far as I can tell, this is the only thing broken about RA2 on bleed at the moment.


alercah:
I somehow missed #291.

PR #294, opened 2016-09-05
New build [experimental]

alercah:
This is an _experimental_ new build approach that I built up. At the moment, this is _not_ suitable for merge, because it is Linux-only and needs to be adjusted for OSX and ported to Windows.

The first part of the approach is to make OpenRA into a submodule of the project. This kind of hurts, because submodules totally suck. But the other options available are:
- Use subtrees. This would work very similarly to submodules without most of the suck. However, there is one very important "feature" of subtrees which makes them unsuitable: `git rebase` does not understand them. You basically cannot rebase through or over a subtree commit, as the rebasing algorithm will not remap the subtree merge correctly. This makes it pretty awful, given the OpenRA rebase-heavy workflow. It might still be workable; if that is the case, then the only thing that needs doing is to switch from submodules to subtrees with nothing further ado. I'm very worried about this, though.
- Do what we currently do, which is keep OpenRA in another directory. I don't like this approach because it necessitates a fragile build system. There's a lot of irritating manual copying of DLLs right now, and cake is basically unportable. Even if we automated some of it, we're still creating a very fragile process .

I wanted to take the approach where we put OpenRA into a subfolder of some kind for the following reasons:
- We can set the csproj to project-reference the OpenRA projects rather only referencing the DLLs. This makes development far more pleasant.
- The dependencies live at fixed & known locations, so we are free to bake these into our files. We would only have to worry about a reorganization of the OpenRA repo.
- Recursive Make is an easy way to do necessary build steps in OpenRA.
- We can correctly & automatically set the version dependencies of our module based on a fixed & known OpenRA commit.

For packaging purposes, I moved all of the mod files into a `mod/` subdirectory. The packaging script in the Makefile currently packages all files in `mod/` that aren't gitignored. `mod.yaml` has its variables substituted when this is done, and the DLL is thrown in to complete the package.

The net result is that all one needs to do to get RA2 running is as follows:
- `git clone` the repository
- `git submodule init`
- `make install`

This builds `ra2.oramod` and installs it. Actually getting OpenRA running must currently be done locally, rather than in an install, because TS is not installed, so that amounts to:
- `cd OpenRA`
- `make`
- `./OpenRA.Game.exe`

(`make dependencies` is invoked by the top-level `make` by default)

Thanks to @Phrohdoh for pointing out some obvious mistakes and helping me with xbuild.


pchote:
You can avoid moving the ra2 files into a mod subdirectory and copying oramods about by adding a `make run` rule that runs `./OpenRA/OpenRA.Game.exe Game.Mod=$(pwd)`

abcdefg30:
Please do that. I don't like the idea of moving all the files into a `mod` subdirectory at all.

alercah:
That could be undone, but then to package we'd have to be more selective and explicitly list every single packageable file. To be fair, though, what I'm doing now is a tad hacky.

Phrohdoh:
Would it be of any help if the utility learned to package a mod into an oramod by reading the manifest and zipping those directories up recursively?

alercah:
Yeah, then the only issue would be the versioning in `mod.yaml`. We could do it the way OpenRA does it, and just leave the file dirty, though.

pchote:
`zip` accepts a `-x` or `--exclude` argument to blacklist files/directories.

alercah:
I'm currently using gitignore, because there are many files, backups, etc. which might get generated which need to be ignored. A whitelist is far less susceptible to accidentally adding a new build script and including that, and will break more obviously, so if `mod/` isn't separated out, I think it's better. Using the `mod.yaml` manifest is probably the best idea, though.

pchote:
Those same additional files (backups etc) would occur with a `mod` subdir, so that shouldn't impact much.

alercah:
That's why I'm currently bypassing any files ignored by git.

Phrohdoh:
> We could do it the way OpenRA does it, and just leave the file dirty, though.

I do not understand this part. It is only dirty in the eyes of git.
If we replace the {DEV_VERSION} with whatever the current version is then do the packaging we should be good to go.

alercah:
That, and Make can't process the dependency correctly.

Phrohdoh:
@abcdefg30 @Mailaender Are either of you interested in taking this?

alercah:
I haven't totally abandoned it. Just been busy. :)

Phrohdoh:
@alercah I understand, but I didn't want to ask you to keep working on it or continue brainstorming if there is no interest from the maintainers.

Mailaender:
This hard-codes the directory so that everyone is forced to use it this way. You can't then just git checkout into `$HOME/.openra/mods/` or `OpenRA\Support\mods` which most people here prefer, but this was already said 19 days ago by @pchote and @abcdefg30.

See also https://github.com/OpenRA2 where I tried essentially the same, but in a cleaner, future proof, split out "make everyone happy" way which ironically satisfied no one and was therefore abandoned.

abcdefg30:
Superseded by #386. Sorry.

PR #298, opened 2016-09-17
Polished the Prism tower laser a bit

Mailaender:
Followup of https://github.com/OpenRA/OpenRA/pull/11912. I took https://youtu.be/ZQWODlIf8eA?t=1m as a reference.

### DIFF
--- a/weapons.yaml
+++ b/weapons.yaml
@@ -1055,8 +1055,8 @@ PrismShot:
 	Range: 8c0
 	Report: bpriat1a.wav
 	Projectile: LaserZap
-		Width: 70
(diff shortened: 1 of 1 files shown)


GraionDilach:
This does use PlayerColor though, except that the laser drawing code brightens the color up a ton - bown becomes full orange even.

pchote:
Related to https://github.com/OpenRA/OpenRA/issues/10064.  When somebody gets around to implementing the PalettedLineRenderer then you could use a custom colour remap to generate the extra-bright colours needed here.

abcdefg30:
Looks like this is not doable yet.

PR #305, opened 2016-09-30
Update the ResourceType trait definitions

abcdefg30:
Updates according to https://github.com/OpenRA/OpenRA/pull/12103.


Mailaender:
Thanks!

PR #315, opened 2016-11-03
Add the UnitCommand widget to ingame-player.yaml

abcdefg30:
See https://github.com/OpenRA/OpenRA/pull/12021.


Phrohdoh:
If you know this is right I'd say just push to master.
RA2 has so little manpower currently that reviews are likely to never happen.

PR #345, opened 2016-12-30
Parameter Fixed

cookgreen:
Fix the parameter type error for Game.Sound.Play function, because the new OpenRA has changed the parameter type to enum SoundType  not string

### DIFF
--- a/OpenRA.Mods.RA2/Activities/ChronoResourceTeleport.cs
+++ b/OpenRA.Mods.RA2/Activities/ChronoResourceTeleport.cs
@@ -37,7 +37,7 @@ namespace OpenRA.Mods.RA2.Activities
 				self.World.AddFrameEndTask(w => w.Add(new SpriteEffect(sourcepos, w, image, info.WarpInSequence, info.Palette)));
 
 			if (info.WarpInSound != null)
-				Game.Sound.Play(info.WarpInSound, self.CenterPosition);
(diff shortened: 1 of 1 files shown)


GraionDilach:
This is flatout wrong. For starters, you break the formatting with replacing tabs with spaces. For the second aspect, The result might compile but I'm certain it'll crash because it wants to parse the enum's type from the enum's internal definition?! (I don't even understand what you wrote, it's that wrong).

The argument should be SoundType.World, reading the commit/PR which introduced this would've told you the same.

cookgreen:
It is useless if the code can't be complied correctly, the code gives two wrong argument, it can't be complied!!! Do you really understand what "can't be compiled" mean??? You can close this pull request(**i don't think it can pass, it is a just warning**), but you guys should update the code and fix the argument type error immediately.

cookgreen:
And by the way, even the code may crash, i don't think it is worse than it can't be compiled correctly, at last i make the code can be compiled correctly

EngineGuy:
Look at pull request #344. It solves the same problem, and I think that it is a correct way of fixing it.
Currently you can compile code using changes from that pull request. Or you can wait until it gets fixed in master branch.

**This** pull request should be closed or fixed.

cookgreen:
I have compiled the code with my change, it works well

GraionDilach:
Closing this in favor of #344.

Issue #346, opened 2016-12-31
Can't locate type 'UpgradeManagerInfo'


cookgreen:
`Exception of type `System.InvalidOperationException`: Cannot locate type: UpgradeManagerInfo`

MustaphaTR:
Upgrades are removed and replaced by conditions on upstream. RA2 mod has not upgraded for it yet.

cookgreen:
but how to fix it? where to find the 'upgrademanagerinfo'? i try to search it in RA2 source code but failed

MustaphaTR:
It is bot anything to do with C# code. It is about .yaml. UpgradeManagerInfo is replaced as i said.

cookgreen:
And i debug the OpenRA source code, and still can't find.
I have found the problem is caused in Ruleset Function(OpenRA), the property 'DefaultRules' of ModData instance - moddata cased this error.

MustaphaTR:
You can't find it because it no longer exists.

cookgreen:
did you mean mod.yaml? i searched, but no results

MustaphaTR:
No, .yaml files under rules folder and UpgradeManagerInfo is not really what you need to search, give me a sec to find the PR about it.

cookgreen:
But which gives the openra this wrong parameter? something like yaml?

MustaphaTR:
Tags that are removed/replaced.

cookgreen:
okay, so how did we fix it?

MustaphaTR:
https://github.com/OpenRA/OpenRA/pull/12497/files

The tag `UpgradeManager:` is renamed to `ConditionManager:` that is the reason of error.

cookgreen:
I have fixed this error and create a pull request

GraionDilach:
Reverse duplicate of #354.

PR #347, opened 2017-01-01
Rename UpgradeManager to ConditionManager

### LINKED ISSUE #354: RA2 is outdated against bleed OpenRA.
I've been struggling for a few hours today to find a good commit in the openra repo (in both the `master` and `bleed` branches) which fully works with the ra2 mod, but there's always some problems.

I've had the most success with some commits around September 2016 where the solution with ra2 will build and run, and the ra2 mod begins to start, indeed it plays the RA2 background music, but there's a bunch of exceptions that happen during type-loading and processing the rules yaml where it fails. I don't want to log bugs because these issues are from targeting much older commits anyway.

The current state of the ra2 project is incompatible with the latest `bleed` commits because the yaml rules have changed considerably and so it needs upgrading, but I'm not familiar with the process, and doubtless the maintainers of this mod are aware of that.

Can the maintainers, or anyone else in the know, advise us which of the most recent openra commits are known to work and run ra2 successfully?

cookgreen:
Update the yaml file for OpenRA bleed. Closes https://github.com/OpenRA/ra2/issues/346. Fixes #354.


rules/animals.yaml
```diff
@@ -1,5 +1,4 @@
 cow:
-	Inherits: ^Animal
```
Mailaender: You shouldn't remove this.

rules/civilians.yaml
```diff
@@ -1,98 +1,98 @@
 civ1:
-	Inherits: ^CivilianInfantry
+	Inherits: ^CivInfantry
```
Mailaender: You are introducing new breakages here:
> OpenRA.YamlException: ra2|rules/civilians.yaml:2: Parent type `^CivInfantry` not found

rules/soviet-infantry.yaml
```diff
@@ -1,6 +1,5 @@
 e2:
 	Inherits@1: ^Infantry
-	Inherits@2: ^Parachutable
```
Mailaender: You shouldn't remove this, either.

rules/allied-infantry.yaml
```diff
@@ -76,13 +75,11 @@ dog:
 		VoiceSet: DogVoice
 	ProducibleWithLevel:
 		Prerequisites: barracks.infiltrated
-	-TakeCover:
```
Mailaender: https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Mods.Common/Traits/Infantry/TakeCover.cs is still available and shouldn't be removed for the dog.

Mailaender:
It seems you are having trouble manually editing the rules. Try using the automated utility commands which are explained at https://github.com/OpenRA/OpenRA/wiki/Upgrading-mods-to-a-new-release

cookgreen:
done

cookgreen:
done

GraionDilach:
Pretty sure we have more issues than that. Someone with time (I don't have such atm though) has to go through all yamls and manu8ally update them.

pchote:
Why aren't you guys using the automated upgrade rules?  This is what they're there for.

GraionDilach:
They were used when this was actively maintained. I stepped down from maintainer status when the cake dependency got in and the others apparently fallen out. I actually lost interest so much that I am thinking of reverting https://github.com/GraionDilach/OpenRA.Mods.AS/issues/2 even.

cookgreen:
@pchote i tried to upgrade it, but the upgrade tool crashed(You can check issues #352)

drifthat:
I guess all the yaml files under rules should be converted as well?

GraionDilach:
Yes, a partial "fix" makes no sense and just confuses people.

m3ntalsp00n:
@cookgreen I've got a version that is now working - doesn't crash and runs the same as before as far as I can tell. 

Following from these discussions, there are 'a lot' of yaml changes and the utility upgrade tool can barf all over it if you're not careful. Have to do a manual update, then run the tools to pick up what you've missed. Will need to update that tool for ra2 - but for now just wanna play :)

Would you like me to commit to your repo or just create a new pull request?

cookgreen:
I am glad that if you can commit my repo

Mailaender:
Thanks. I squashed everything together as you introduced some junky merge commits.

abcdefg30:
Thanks for the cleanup! I pushed two minor follow-up commits and will hopefully have some more time to spend on this project this weekend.

PR #355, opened 2017-01-17
Split content metadata into their own files

drifthat:
to match https://github.com/OpenRA/OpenRA/pull/11771.


drifthat:
CI build is passed now

drifthat:
merged w/ latest master codebase

GraionDilach:
This needs a rebase.

Mailaender:
It looks like the GitHub desktop clients makes things worse by merging in the latest master instead of just rebasing, but I could resolve this with the web interface by squashing things together also eliminating the revert commit which then also let's us avoid a commit that breaks things.

GraionDilach:
I can't tell any desktop GUI client which rebases on default settings, so that doesn't apply to Github's alone.

PR #359, opened 2017-02-04
Rename 'UpgradeOverlay' because of the new version of OpenRA

cookgreen:


rules/defaults.yaml
```diff
@@ -83,8 +83,6 @@
 		Conditions: crate-firepower, crate-damage, crate-speed
 
 ^IronCurtainable:
-	UpgradeOverlay@ironcurtain:
```
m3ntalsp00n: Should be renamed to WithColoredOverlay not removed entirely.

cookgreen: okay

cookgreen: changed

cookgreen:
Fixed

PR #367, opened 2017-02-14
Abuse 'ls' to determine if content directory is empty

Phrohdoh:
The directory existing is completely valid.
We want to know whether or not the directory is _empty_.

### DIFF
--- a/fetch-content.sh
+++ b/fetch-content.sh
@@ -9,7 +9,7 @@ else
 	DIR="$HOME/.openra/Content/ra2/"
 fi
 
-if [ ! -e "$DIR" ]; then
+if [ -z "$(ls -A "$DIR")" ]; then
 	echo "Downloading RA2 mod content"
 
 	mkdir -p "$DIR"


fetch-content.sh
```diff
@@ -9,7 +9,7 @@ else
 	DIR="$HOME/.openra/Content/ra2/"
 fi
 
-if [ ! -e "$DIR" ]; then
+if [ -z "$(ls -A "$DIR")" ]; then
```
samrocketman: If directory doesn't exist then it will output to stderr.  Perhaps,

```bash
if [ ! -e "$DIR" -o -z "$(ls -A "$DIR")" ]; then
...
```
or discarding stderr of `ls -A`.

samrocketman: Also, regarding validity; it would probably be better to test for a specific expected file rather than checking for any arbitrary file with `ls -A`.

abcdefg30:
Superseded by #385.

PR #368, opened 2017-02-14
Fix an indexing issue in build.cake

Phrohdoh:
This depends on #367 because you want that too. :-)

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -10,8 +10,8 @@ var configuration = Argument("configuration", "Debug");
 var rootDir = Directory(".");
 var depsDir = Directory("./OpenRA.Mods.RA2/dependencies");
 
-// TODO: Combine 'deps' and 'depsInOpenRA' into an array of pairs/structs
-var deps = new[] {
+// TODO: Combine 'depFilenamesInRA2DepsDir' and 'depsInOpenRA' into an array of pairs/structs
+var depFilenamesInRA2DepsDir = new[] {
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -69,31 +69,39 @@ Task("deps").Does(() => {
     for (var i = 0; i < missingDepsCopy.Length; i++)
     {
         var dep = missingDepsCopy[i];
-        var depPathInOpenRA = depsInOpenRA[i];
+        var depPathInOpenRA = depsInOpenRA.SingleOrDefault(d => d.EndsWith(dep));
+        if (depPathInOpenRA == null)
+            throw new Exception(nameof(depsInOpenRA) + " does not contain an entry for " + dep);
```
Phrohdoh: @abcdefg30 Please let me know if this fails to run for you.

I can remove the `nameof` easily I was just being lazy.

GraionDilach:
The issue with cake is that it is included so I don't see this as a fix.

Phrohdoh:
@GraionDilach You are more than welcome to submit a cross-platform solution to fetching dependencies, building, and cleaning.

GraionDilach:
There was a solution which you prominently overwrote, along with my efforts to actually contribute into this repository.

abcdefg30:
#386 was merged and `build.cake` removed. We could still use the first commit though.

abcdefg30:
No activity.

PR #369, opened 2017-02-14
Set the last known good engine version to eef4c9b62

Phrohdoh:


Phrohdoh:
This feature will not be part of OpenRA much longer.

PR #377, opened 2017-04-02
Adds faction descriptions

colstrom:
I noticed these were missing, so I added them. As far as I'm aware, these are the only notable differences between the factions. I opted for more specific unit descriptions (Special Infantry/Vehicle/Aircraft/Building) over the generic "Special Unit" used in other mods. This seems clearer for those not already familiar with the units.


colstrom:
Thanks!

Issue #361, opened 2017-02-05
Aircraft fly off screen when built


ibrokemypie:
Both harriers and black eagles spawn from the airfield then immediately fly off screen to the left and disappear, but still fill up said airfield. This makes them useless and unusable.

rob-v:
Fixed by PR #381

abcdefg30:
Actually, it doesn't really fly off the screen, just takes a very long turn to land on the airfield again.

rob-v:
As I understand it: correct behaviour for planes vs helis is achieved only by CanHover flag which determines if it is Plane or Heli. The planes fly off the screen and never return. If I change it to CanHover: False, it means they are considered Planes instead of Helis, they properly return. Without this fix, they don't return and behave in general wrong.

GraionDilach:
Not exactly wrong.

Thing is that RA2 planes used a mixture of the OpenRA logics. While yes, planes were VTOL as helicopters, the OpenRA helicopter logic climbs cliffs vertically (doing a Z move pretty much), exactly as RA2 jumpjets (Rocketeer, Kirov, Siege Chopper, Blackhawk), RA2 planes passed cliffs as smoothly as the OpenRA plane logic.

Ideally, the RA2 planes would only use heli logic between landing and taking off, with switching to plane logic immediately after reaching CruiseAltitude.

rob-v:
Thanks for explanation. It seems that currently OpenRA supports either planes (CanHover=false) or helis (CanHover=true). In RA2 is wanted mixture of logics, which is currently not supported. In my opinion to have useable/playable aircraft in RA2 until it is enhanced it is needed to make RA2 aircraft also either OpenRA planes or helis, not mixture atm.

Mailaender:
They don't do this anymore. Instead they take off and immediately land again.

PR #381, opened 2017-04-13
Fix Aircraft fly off screen when built #361

rob-v:


rob-v:
IMO Harrier(orca) and Black Eagle(beag) are planes, not helicopters so to work properly they need to be recognized as planes thanks to CanHover=False.
IsPlane = !info.CanHover;

PR #382, opened 2017-04-28
Add alternative download methods to fetch-content.sh

martinsik:
I was testing RA2 on OS X and I found out that the download script `fetch-content.sh` relies on `wget` that isn't a standard part of OS X so I added an alternative with `curl`. Also, OS X has `bash 3.2.*` so I couldn't simply use associative arrays.

However, it's still not completely bulletproof, if you don't have `7z` on your system it'll download the files and never retry the `7z` command because of `if [ ! -e "$DIR" ]; then`.

### DIFF
--- a/fetch-content.sh
+++ b/fetch-content.sh
@@ -2,6 +2,7 @@
 
 # Die on any error for Travis CI to automatically retry:
 set -e
+
 if [ "$(uname)" == "Darwin" ]; then
 	DIR="$HOME/Library/Application Support/OpenRA/Content/ra2/"
 else


abcdefg30:
I'm not sure if `fetch-content.sh` was ever designed for anything else than travis, but it still seems to work there. I don't know much about this, so it would be good if @Mailaender or @Phrohdoh could have a look.

Mailaender:
You can use `which` to check of `curl` or `wget` exists to simplify this.

martinsik:
@Mailaender Using `which` is not recommended. http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

Phrohdoh:
Just use `uname` to determine the host OS and run the necessary commands.
No need to make this complex with replacements and command building+eval.

@abcdefg30 Is right in that this was never intended as a way for individuals to download the content files. This is only for Travis-CI.

fetch-content.sh
```diff
@@ -15,7 +15,28 @@ if [ ! -e "$DIR" ]; then
 	mkdir -p "$DIR"
 	pushd "$DIR"
 
-	wget http://xwis.net/downloads/Red-Alert-2-Multiplayer.exe
+	DOWNLOADERS=("wget" "curl")
+	DOWNLOAD_COMMANDS=("wget URL" "curl -o \"Red-Alert-2-Multiplayer.exe\" URL")
+
+	for INDEX in "${!DOWNLOADERS[@]}"; do
+		DOWNLOADER=${DOWNLOADERS[$INDEX]}
```
samrocketman: Unsafe unquoted variable.

fetch-content.sh
```diff
@@ -15,7 +15,28 @@ if [ ! -e "$DIR" ]; then
...
+		EXISTS=1
+		(hash ${DOWNLOADER} 2> /dev/null) || EXISTS=0
+
+		if [ $EXISTS -eq 1 ]; then
+			COMMAND=${DOWNLOAD_COMMANDS[$INDEX]}
```
samrocketman: Unsafe unquoted variable.

samrocketman:
Use `type -p <program>` instead of `which`.

martinsik:
@Phrohdoh @abcdefg30 I was following the instructions here https://github.com/OpenRA/ra2/wiki/Step-by-Step#unix that mention that I need to run this script. If I'm not supposed to use this script is there some other way to get the original resources?

@samrocketman Can you be more specific? Btw, I'm not using `which`.

Iamgoofball:
If you've got a copy of RA2 on disk or the origin complete pack of all C&C games, you can pull the files from those.

Iamgoofball:
Or, you could install the XWIS version manually and pull the files from it.

abcdefg30:
> @Phrohdoh @abcdefg30 I was following the instructions here https://github.com/OpenRA/ra2/wiki/Step-by-Step#unix that mention that I need to run this script. If I'm not supposed to use this is there some other way to get the original resources?

Ah, right. I updated the instructions slightly. You usually get the game content from original CDs using the ingame content installer (like @Iamgoofball said).

I guess we could still take this if you can address the review comments.

samrocketman:
@martinsik can you elaborate on what you'd like me to be more specific about.  My comment on `type` vs `which` was in response to https://github.com/OpenRA/ra2/pull/382#issuecomment-298162755 ; I agree which is not recommended and offered a shell built-in.

If you're talking about my earlier comment.  I don't recommend using bash parameter expansion and arrays to mix how a tool gets downloaded by using multiple arrays.  There are simpler ways to do what you're doing.  For example,

```bash
#!/bin/bash

set -e

function download_url() (
  set -e
  URL="$1"
  LOCAL_FILE="${1##*/}"
  if [ -x "$(type -p curl)" ]; then
    curl -f -o "${LOCAL_FILE}" "${URL}"
  elif [ -x "$(type -p wget)" ]; then
    wget "${URL}"
  else
    echo "No supported download method found." 1>&2
    return 1
  fi
)

if [ "$(uname)" == "Darwin" ]; then
  DIR="$HOME/Library/Application Support/OpenRA/Content/ra2/"
else
  DIR="$HOME/.openra/Content/ra2/"
fi

[ -d "${DIR}" ] || mkdir -p "${DIR}"

pushd "${DIR}"

#shell will automatically exit if returns non-zero because of set -e
download_url "http://xwis.net/downloads/Red-Alert-2-Multiplayer.exe"

7z e Red-Alert-2-Multiplayer.exe
rm *.exe *.dll *.DLL *.wav *.mmp *.CFG *.WAR *.cache
popd
```

Note: without `curl -f` curl will return a SUCCESS exit code even with 404 HTTP responses.

samrocketman:
Since I went through the effort of explaining what I meant I decided to just open a PR.

fetch-content.sh
```diff
@@ -15,7 +15,28 @@ if [ ! -e "$DIR" ]; then
 	mkdir -p "$DIR"
 	pushd "$DIR"
 
-	wget http://xwis.net/downloads/Red-Alert-2-Multiplayer.exe
+	DOWNLOADERS=("wget" "curl")
+	DOWNLOAD_COMMANDS=("wget URL" "curl -o \"Red-Alert-2-Multiplayer.exe\" URL")
```
samrocketman: Using `curl` this way will never show a failure.  `curl` will not surface a nonzero exit code if a file does not exist.  Example:

```
curl -o /dev/null http://example.com/does-not-exist.tar
echo $?
curl -f -o /dev/null http://example.com/does-not-exist.tar
echo $?
```

abcdefg30:
@martinsik what is the status here?

abcdefg30:
No activity since a while. Feel free to reopen this if you got time to update it.

PR #385, opened 2017-05-07
Fall back to curl if wget is missing

samrocketman:
Supersedes #382.  

Bugfixes which were bugs in the old script:

- Mac OS doesn't come with `wget` by default, but it does come with `curl`.
- If the ra2 content directory existed, then the script would skip attempting to download even if there was no content.  Fixes `travis_retry ~/.openra/mods/ra2/fetch-content.sh` in `.travis.yml`.
- If `wget` was attempted more than once and the download file already exists, then `wget` will download `Red-Alert-2-Multiplayer.exe.1` and so on.  `wget -O` forces download to `Red-Alert-2-Multiplayer.exe`.

### DIFF
--- a/fetch-content.sh
+++ b/fetch-content.sh
@@ -3,20 +3,36 @@
 # Die on any error for Travis CI to automatically retry:
 set -e
 
+function download_url() {
+  URL="$1"
+  LOCAL_FILE="${1##*/}"
+  if [ -x "$(type -p wget)" ]; then
+    wget -O "${LOCAL_FILE}" "${URL}"
+  elif [ -x "$(type -p curl)" ]; then
+    curl -f -o "${LOCAL_FILE}" "${URL}"
+  else
+    echo "No supported download method found." 1>&2
+    return 1
+  fi
+}
+
 if [ "$(uname)" == "Darwin" ]; then
-	DIR="$HOME/Library/Application Support/OpenRA/Content/ra2/"
+  DIR="$HOME/Library/Application Support/OpenRA/Content/ra2"
 else
-	DIR="$HOME/.openra/Content/ra2/"
+  DIR="$HOME/.openra/Content/ra2"
 fi
 
-if [ ! -e "$DIR" ]; then
-	echo "Downloading RA2 mod content"
-
-	mkdir -p "$DIR"
-	pushd "$DIR"
-
-	wget http://xwis.net/downloads/Red-Alert-2-Multiplayer.exe
-	7z e Red-Alert-2-Multiplayer.exe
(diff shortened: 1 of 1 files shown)


fetch-content.sh
```diff
@@ -3,20 +3,36 @@
...
+  cd "${DIR}"
+  #only exit if there's existing content
+  if ls *.mix &> /dev/null; then
+    exit 0
+  fi
+}
+echo "Downloading RA2 mod content"
+cd "${DIR}"
+#download the file else exit non-zero
+download_url "http://xwis.net/downloads/Red-Alert-2-Multiplayer.exe"
```
Phrohdoh: This `return`s `1` but doesn't actually `exit`.

samrocketman: It exits because of `set -e`.  `return 1` is considered a non-zero exit code.  Refer to [bash manual](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html).  I also manually test this by modifying `$PATH`.

samrocketman: I wrote a quick script so that you can test an example yourself.  https://gist.github.com/samrocketman/60d307dd8d448a080522346011e4ea41

Phrohdoh: I did not know about `-e`'s exiting on a non-zero return, TIL.

samrocketman:
Keep in mind this legitimately fixes a bug for Travis CI with `travis_wait` which doesn't work with the script in master.

samrocketman:
What is left for me to do for this change to be considered good enough to merge?

samrocketman:
Actually, I'd like to ammend this for `type -P` instead of `type -p`.

fetch-content.sh
```diff
@@ -3,20 +3,36 @@
 # Die on any error for Travis CI to automatically retry:
 set -e
 
+function download_url() {
+  URL="$1"
+  LOCAL_FILE="${1##*/}"
+  if [ -x "$(type -p wget)" ]; then
```
samrocketman: Use `type -P` instead, because it is more reliable.

ref: `type` docs https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html

>If the `-p` option is used, `type` either returns the name of the disk file that would be executed, or nothing if `-t` would not return `'file'`.
>
>The `-P` option forces a path search for each _name_, even if `-t` would not return `'file'`.

Mailaender: I assume you don't want this merged yet then.

samrocketman: Doesn't really matter to me.  If you wanted to merge it then you know what to chamge.

PR #386, opened 2017-05-07
Adapt the mod to the new mod layout

abcdefg30:
This reintroduces the OpenRA/OpenRA submodule and comes with the new scripts from
https://github.com/OpenRA/OpenRAModTemplate. I still need to update some of the scripts (line endings, cake, travis), but the general structuring is finished.


Mailaender:
I assume you need to update https://github.com/OpenRA/ra2/blob/master/.travis.yml#L24 as well with the new paths. Otherwise looks sane on a first look.

OpenRA.Mods.RA2/OpenRA.Mods.RA2.csproj
```diff
@@ -43,16 +43,16 @@
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="OpenRA.Game">
-      <HintPath>dependencies\OpenRA.Game.exe</HintPath>
+      <HintPath>..\engine\OpenRA.Game.exe</HintPath>
```
pchote: If you change these to project references then you'll be able to use the "Jump to Definition" IDE feature.

build.cake
```diff
@@ -1,119 +1,32 @@
...
-//
-// Example:
-// cake -auto=true -openra-root=~/projects/openra
 
 var target = Argument("target", "default").ToLowerInvariant();
 var configuration = Argument("configuration", "Debug");
 
-var rootDir = Directory(".");
-var depsDir = Directory("./OpenRA.Mods.RA2/dependencies");
+Task("default").Does(() => {
```
pchote: In the longer term we should implement this functionality in the mod template (using the standard makefile / batch / powershell scripts) so that all mods can benefit.

launch-game.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+TEMPLATE_LAUNCHER=$(python -c "import os; print(os.path.realpath('$0'))")
+TEMPLATE_ROOT=$(dirname "${TEMPLATE_LAUNCHER}")
+source ${TEMPLATE_ROOT}/mod.config
+
+MOD_SEARCH_PATHS="./../.."
```
pchote: This sets the mod search path to the parent directory of the RA2 clone.  This means that if you have a shared source directory (e.g. `~/src/` with subdirectories for e.g. `ra2` and `d2`) then this will pick up the other mods as well as ra2.  This will probably break things in unexpected ways.

pchote: As painful as it is to break the git history, the best fix here is to move the mod files into a `mods/ra2` directory as done in the template.

mod.config
```diff
@@ -0,0 +1,8 @@
+# The id of the mod to launch by default.  This must exist as a directory in the mods directory
+MOD_ID="ra2"
+
+# Import the default OpenRA mods for use as external mod references
+# in your mod.yaml. Accepts values "True" or "False".
+# WARNING: This setting is provided to simplify early mod development
+# and must be disabled before installers can be generated
+INCLUDE_DEFAULT_MODS="True"
```
pchote: You should aim to make whatever fixes you need to reset this to "False" in the shortish term.

abcdefg30:
This waits for https://github.com/OpenRA/OpenRAModTemplate/pull/3 for now.

launch-game.sh
Iamgoofball: I'm fairly certain that GitHub retains history when it detects a file move.

pchote: It doesn't.  See e.g. https://github.com/OpenRA/OpenRA/commits/bleed/OpenRA.Mods.Cnc/Traits/Chronoshiftable.cs.

GraionDilach: It's weird - blame definitely retains the history. https://github.com/OpenRA/OpenRA/blame/bleed/OpenRA.Mods.Cnc/Traits/Chronoshiftable.cs

samrocketman: It depends.  Git doesn't track files, per se, but blobs of text.  The history will be retained if, in a single commit, only a rename happens.  However, if you rename and change code in the same commit then it depends on how much you change.  In general renaming files and folders should be kept as its own commit.

abcdefg30:
Updated. It should work now, but doesn't as it can't find the `cnc.dll` for some reason.

pchote:
Your problem is that you have a stray `$cnc: cnc` in your package list (you don't use it?), and `INCLUDE_DEFAULT_MODS` doesn't appear to be working so the utility can't find it.

abcdefg30:
Finally found the time to update this again.

pchote:
I have a mostly done PR (linux/macOS done, just needs the windows scripts) for https://github.com/OpenRA/OpenRAModTemplate/issues/7.  If you can wait for that it would save you from having to add and then immediately remove a submodule.

pchote:
https://github.com/OpenRA/OpenRAModTemplate/pull/14 has now been merged.

alercah:
In addition to line endings, the bash scripts need to have +x set.

samrocketman:
@alercah 
> In addition to line endings, the bash scripts need to have +x set.

I looked over all of the bash scripts and none of them explicitly call `set -x` nor pass `-x` via the shabang.  Can you elaborate?  `set +x` is the default for `/bin/sh` and `/bin/bash`.

alercah:
I mean the executable bit, not the bash setting.

samrocketman:
Cool,

    find . -type f -name '*.sh' -exec chmod 755 {} +

alercah:
Sure, that should be in this patch though.

abcdefg30:
Finally fixed & updated.

abcdefg30:
Forgot to set `INCLUDE_DEFAULT_MODS` to `False`, fixed.

abcdefg30:
Fixed travis/linux.

PR #390, opened 2017-08-09
Fetch engine windows

hartmark:
Added bat file for fetching the latest OpenRA engine.

Also if you want to debug from visual studio, you need to do the following:
1. Set OpenRA.Game as startup project
2. Edit project properties and set Command Line Arguments to this:
`Game.Mod=ra2 Engine.ModSearchPaths=.\mods,.\..\mods`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ temp.s
 # SublimeText
 *.sublime-project
 *.sublime-workspace
+/.vs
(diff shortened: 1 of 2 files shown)


abcdefg30:
Why doesn't the normal `make.cmd`/`make.ps1` work for you sufficient (i.e. what needs to be changed there)? This also ignores the engine version set in `mod.config`.

hartmark:
Man I feel stupid, I thought that the make.cmd just was for compiling.

And the "make.cmd all" works perfectly to download the engine and all dependancies.

I'm closing this PR

abcdefg30:
Ah, maybe we should make that more clear on the wiki page.

PR #391, opened 2017-08-19
Remove unused private field

stevenpost:
This generates a compile warning, which is treated as an error when invoking 'make' on Linux.
Thus making the mod able to build again.


abcdefg30:
Thanks.

Issue #189, opened 2016-04-15
GI not able to deploy


Elthan:
GIs are not able to deploy like they can in the original game:

![ra2_gi](https://cloud.githubusercontent.com/assets/7212361/14570578/57cc58da-0345-11e6-9bf6-aa5472e77e23.PNG)

GraionDilach:
Yeah, it seems troublesome at this stage.

Mailaender:
See https://github.com/OpenRA/ra2/pull/168 for a failed attempt.

Elthan:
I see. Is the idea to only do it with editing the YAML file or is it possible to (f.ex.) create a custom upgrade to do it?

Phrohdoh:
Upgrades via yaml.

ibrokemypie:
+1

forcecore:
With condition system, I think this is becoming possible.
Since GIs will have deployed and undeployed WithInfantryBody, then only WithMakeAnimation needs to be used, as thumper does, hopefully.

https://github.com/OpenRA/OpenRA/pull/13737

GraionDilach:
The main problem here is multiple turrets support or something along the way. TakeCover uses a turret and the deployed GI should also be a "turret".

forcecore:
Oh yeah I remember, it's been ages ago since I implemented fire ant.

Mailaender:
I made a new attempt https://github.com/OpenRA/ra2/pull/394 to find out the actual status of the feature.

Mailaender:
https://github.com/OpenRA/ra2/pull/547

PR #394, opened 2017-08-24
Added GI sandbag deployment

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/189.

It still has some limitations. All of them are engine related:

- [ ] <s>`Crushable` needs to become conditionable.</s> OpenRA/OpenRA#14507
- [ ] `TakeCover` needs to become conditionable.
- [x] `WithInfantryBody` won't aim in the right direction. OpenRA/OpenRA#12738

I also experimented with https://github.com/OpenRA/OpenRA/pull/13736, https://github.com/OpenRA/OpenRA/pull/13143 and https://github.com/OpenRA/OpenRA/pull/13745, but couldn't get them to work. This will need several followups.


MustaphaTR:
I'm pretty sure deployed GIs are still crushable in original(at least in YR), but Guardian GIs aren't in YR. Not sure what control these in original games ini code tho.

GraionDilach:
>I'm pretty sure deployed GIs are still crushable in original(at least in YR), but Guardian GIs aren't in YR.

Exactly. http://modenc.renegadeprojects.com/DeployedCrushable was only add for the GGI.

ABrandau:
Hm, nice to see you are back modding, this reminds me I had to go for a silly hack to get quasi GI's working on my mod. It was an Attack turreted with an invisible turret, and the firing animation was done through a muzzle anim, hope you can get it done better now.

Mailaender:
Removed the `Crushable` TODO.

I also experimented with the muzzle, which will get the firing directions right, but not the player palette or the aiming. Didn't include it, as that won't be the proper way to fix it. The sandbag isn't just a muzzle, but a all-in-one sprite.

Mailaender:
Updated with the `TurnsWhileImmobile` trait from https://github.com/OpenRA/OpenRA/pull/13916 by  @forcecore.

abcdefg30:
Needs a rebase. Unfortunately we can't immediately update the engine version after the fix for OpenRA/OpenRA#12738 as we still depend on `UnitCommandWidget` which was removed upstream.

Mailaender:
This needs to wait for https://github.com/OpenRA/OpenRA/pull/13916 to be actually commited. I assume the easiest way for `UnitCommandWidget` would be to copy a temporary of the old logic until @x-a-n-a-x or someone else crafts some UI.

Valkirie:
This was fixed by [#14846](https://github.com/OpenRA/OpenRA/pull/14846). Can you guys move on with this one now ?

Mailaender:
It still depends on https://github.com/OpenRA/OpenRA/issues/13285 and something that allows it to turn while immobile.

MustaphaTR:
Closing this, i have a working GI that i'll PR soon and this is pretty old and requires rebases.

Issue #336, opened 2016-11-18
All ships - missing ripple effect when moving


Fincer:
Affected units - Allies:

- Destroyer
- Aegis Cruiser
- Amphibious Transport

Affected units - Soviet:

- Sea Scorpion
- Amphibious Transport

Short sample video here:

[![RA2 Squid Idle Animation](https://img.youtube.com/vi/TP9G7sFl8pU/mqdefault.jpg)](http://www.youtube.com/watch?v=TP9G7sFl8pU&t=1m32s)

PR #395, opened 2017-08-25
Add water trails to ships

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/336.

The palette is wrong and none of the currently setup ones seems to fit.

### DIFF
--- a/mods/ra2/rules/soviet-naval.yaml
+++ b/mods/ra2/rules/soviet-naval.yaml
@@ -64,6 +64,7 @@ sub:
 	Mobile:
 		TurnSpeed: 4
 		Speed: 71
+	-LeavesTrails:
 	RevealsShroud:
 		Range: 6c0
 	-Targetable:
--- a/mods/ra2/sequences/misc.yaml
+++ b/mods/ra2/sequences/misc.yaml
@@ -277,6 +277,14 @@ moveflsh:
 		Length: *
 		Tick: 30
 
+wake:
+	idle: conquer|wake1
+		Length: *
+		Tick: 180
+		ZOffset: -512
+		ZRamp: 1
+		Offset: 0, 0, 1
+
 resources:
 	Defaults:
 		UseTilesetExtension: true
--- a/mods/ra2/mod.yaml
+++ b/mods/ra2/mod.yaml
@@ -17,7 +17,7 @@ Packages:
 	~audio.mix
 	~cache.mix
 	~cameo.mix
-	~conquer.mix
+	~conquer.mix: conquer
 	~generic.mix
 	~isogen.mix
 	~isosnow.mix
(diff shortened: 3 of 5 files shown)


GraionDilach:
Make it 50% transparent. Translucent=yes was basically a rough fade-out translucent effect (25%-50%-75% steppings IIRC), so indeed there isn't an 1:1 solution.

Mailaender:
Do you happen to know the water palette? I need to use something as base to make translucent.

mods/ra2/rules/soviet-naval.yaml
```diff
@@ -64,6 +64,7 @@ sub:
 	Mobile:
 		TurnSpeed: 4
 		Speed: 71
+	-LeavesTrails:
```
Phrohdoh: Should this be conditional based on whether or not the sub is submerged?

GraionDilach: In RA2, leaving trails had nothing to do with cloaking, but to Underwater=yes, which disabled trails altogether if present (along with some other effects like sub targeting for NavalTargeting modes).

Mailaender:
> Make it 50% transparent. Translucent=yes was basically a rough fade-out translucent effect (25%-50%-75% steppings IIRC), so indeed there isn't an 1:1 solution.

I just read about https://github.com/OpenRA/OpenRA/issues/10727, which is what you want me to fake here.

GraionDilach:
And I just noticed that you were doubted on the actual palette - yes, effect is the correct one. Which is why I ignored the question.

Mailaender:
Alright. Thanks.

I won't add the fade out effect, yet. At least not in this pull request.

Phrohdoh:
Why not if it is part of this feature?

Mailaender:
Personal time constraints and this probably being more of an upstream engine feature rather than exclusive mod code.

GraionDilach:
I'm still voting for effect50 (akin to how it's set up in Dune2000) as a compromise.

Mailaender:
> I'm still voting for effect50 (akin to how it's set up in Dune2000) as a compromise.

Updated and called it `effect50alpha`.

DylanC:
@Mailaender  - Can this be merged in?

Mailaender:
@abcdefg30 is the current maintainer of the ra2 mod. He and the reviewers decide.

mods/ra2/sequences/misc.yaml
```diff
@@ -277,6 +277,14 @@ moveflsh:
 		Length: *
 		Tick: 30
 
+wake:
+	idle: wake2
```
MustaphaTR: This is wrong. Correct file is the wake1.shp under conquer.mix. There are wake1.shps under other mixes too tho so just writing wake1 doesn't work. This needs a setup like TS cameos use.

Mailaender: Corrected that.

PR #415, opened 2018-01-22
Add map The Alamo (2)

MustaphaTR:
Small 2 player urban map.


Mailaender:
The sand pit has no ore:

image

Probably an error in the terrain definitions not allowing ore on there.

Mailaender:
Actually probably solved in https://github.com/OpenRA/ra2/pull/419.

PR #419, opened 2018-01-22
Allow ore to grow on road terrain

MustaphaTR:
Otherwise Spawn A's starting ore is gone on The Alamo


Mailaender:
Instead of touching the automatically generated terrain definitions you should probably edit the `AllowedTerrainTypes` instead. Compare https://github.com/OpenRA/ra2/pull/30/commits/5dbc6f4af5078975938acfb3c22e362ea503b8cb

MustaphaTR:
Updated

PR #420, opened 2018-01-22
Make Crazy Ivan instantly explode on death

MustaphaTR:


MustaphaTR:
I did not left the `Report:` intendedly, but apperantly it does.

PR #422, opened 2018-01-22
Fix Harrier and Black Eage deal no damage

### LINKED ISSUE #362: Black Eagle does no damage
Black eagles attack, but their missiles do not damage the target

MustaphaTR:
Fixes #362.

Spread was too low.


Mailaender:
Looks like this wasn't enough. They still don't deal damage against buildings.

Mailaender:
Nevermind https://github.com/OpenRA/ra2/pull/407 fixed that.

PR #423, opened 2018-01-22
Add Terror Drone

MustaphaTR:
For now just works like attack dog. No vehicle targeting. Mostly added as a placeholder unit.

Also defines Drone and Rocket (those were called special_1 and special_2 on original, no reason to use that name here) armor classes on weapons
And some other weapon fixes i noticed while doing that


Mailaender:
> And some other weapon fixes i noticed while doing that

Can you put those into separate commits?

MustaphaTR:
Updated.

PR #424, opened 2018-01-22
Some SW Related stuff

MustaphaTR:
~~I added `DisplayTimerStances:` but it is not working, not sure why.~~

~~Edit: Related chrome code is missing, i'll add that when i back home.~~


MustaphaTR:
That chrome code fix is steaked in with the Command Bar PR. This should work fine after a rebase.

PR #425, opened 2018-01-23
Some parachute fixes

MustaphaTR:
Makes cargo plane non-repulsable.

Makes paradroped units properly only render paradrop image.

Makes paradropped units properly only target by AA.

Gives ^Parachutible to Yuri and fixes issues with GrantTimedConditionOnDeploy that prevents this.

Adds ^ParachutibleVehicle and gives it to ^Vehicles.


Mailaender:
Also the commit message should be more precise.

MustaphaTR:
I think we should make all units parachutable. In skirmish, you would only see GI and Conscript being paradropped, but there are other stuff in several missions. Not all infantry has paradrop anim, but original game used standing image for such units, if you try to paradrop them.

MustaphaTR:
Updated and added a bit more stuff.

PR #427, opened 2018-01-23
Fix wall shadow frames and add critical damage state

MustaphaTR:


abcdefg30:
This looks somehow a bit off:
image

pchote:
@abcdefg30 I expect that won't be fixable until you enable the depth buffer in this mod.

PR #451, opened 2018-02-17
Add Assault Move and Command Bar

### LINKED ISSUE #364: UI Missing bottom bar
The bottom bar as seen above is missing. This is used to select command/attack modes + formations

MustaphaTR:
Fixes #364

Updates the engine version to https://github.com/OpenRA/OpenRA/commit/049ed08, this shouldn't effect another PR, as my local version with all my PRs in it works fine.


LinuxDonald:
Have openra2 no reviewers? :(

PR #480, opened 2018-05-30
Add Stolen Tech Units

MustaphaTR:
Like Chrono Legionnaries, Chrono Commando and Chrono Ivan currently limited to use Chrono Tank logic and can normally walk.


GraionDilach:
Psi Commando threw a lot of rule errors.

PR #250, opened 2016-06-28
Add more exits to infantry producers

Phrohdoh:
screen shot 2016-06-28 at 9 14 27 pm
screen shot 2016-06-29 at 9 50 55 am

Issue #137, opened 2016-02-13
Can't deploy units from factories if there are obstacles [$5]


Fincer:
This issue is clearly visible here in the following Youtube video:

http://www.youtube.com/watch?v=QrL8m48TePY&t=4m43s

In the video, I was producing a few ships. However, I noticed that the ships couldn't be deployed so I tried to set a waypoint and repeat the building process again. Still no luck. My final conclusion was that I needed to sell the existing Naval Yard and construct it again offshore so that naval units could expectedly be deployed.

When a Naval Yard locates near a shoreline, expected behaviour is that units would be deployed from the furthest corner of the building (or something alike) so that this issue wouldn't exist

Fincer:
Same kind of problem with Cloning Vats:

![cantproduce-openra2](https://cloud.githubusercontent.com/assets/3705231/13030342/c5fad320-d2af-11e5-885c-aa1b242f549f.gif)

Cloned infantry would never be deployed. I sold my ConYard and couldn't see cloned infantry there.

abcdefg30:
That is expected behaviour. If you block all exits, the units can't and won't leave the building.
But I guess e.g. the naval yard could use quite a few more exits...

Fincer:
Well, not exactly. This is how it works in the original RA2:

![cantproduce-ra2-cropped](https://cloud.githubusercontent.com/assets/3705231/13031055/d9c60fa2-d2c6-11e5-9fe1-ac1cee48c85c.gif)

You can clearly see cloned units coming out from the Cloning Vats as well, using just another/opposite direction. This is not the case in OpenRA/ra2.

Icenowy:
As I know, at least War Factories will be not able to produce anything if they're all blocked in original RA2.(I've heard this as some tactics, that if you use vehicles to block all the enemies' war factories, the vehicles in the task queue will just lost, and the money is cost) (But in OpenRA the money won't be cost, only the vehicle task queue will be on hold)

MustaphaTR:
> the vehicles in the task queue will just lost, and the money is cost

Not sure about WFAC but infantry in RA2 that coudn't build gives The money back. There is even a bug that u can wall your barracks and u'll still get a free unit from Cloning Vats, but this is fixed in Ares.dll.

Icenowy:
@MustaphaTR You're right. (This cloning bug seems to be also fixed in NPatch AE)

And my experiment shows that on original RA2 infantry will try to escape the structure from any hole of the wall around the structure.

[...]

PR #487, opened 2018-06-08
Remove wrong exits of Allies Barracks.

hathlife:
Currently the Allies Barracks have 2 wrong exits compared to the original game.This PR fix the problem.


MustaphaTR:
That extra exits are intented and was added in https://github.com/OpenRA/ra2/pull/250 as a fix to https://github.com/OpenRA/ra2/issues/137

hathlife:
I have checked the original game,different from soviet barracks,the exit cell of allies barracks would never be blocked by any building like #137 ,the units can still go out,and the units would never go out from the back.

PR #497, opened 2018-11-01
Add GI Deployment

MustaphaTR:
Removed `TakeCover:` from GI for now, since we need it conditional and not apply when deployed.


mods/ra2/rules/allied-infantry.yaml
```diff
@@ -125,26 +125,64 @@ e1:
...
+		RequiresCondition: !rank-elite && deployed
+	Armament@elite-deployed:
+		Weapon: paraE
+		Name: deployed
+		RequiresCondition: rank-elite && deployed
+	GrantConditionOnDeploy:
+		DeployedCondition: deployed
+		UndeployedCondition: undeployed
+		CanDeployOnRamps: True
+		Facing: 160
```
Phrohdoh: Is there any way for us to get the facing dynamically?

Us forcing it to this value means the sprite comes out of the undeploy sequence and may abruptly face a different direction.

Not a block but could be polished (potentially in the future with upstream changes).

mods/ra2/rules/allied-infantry.yaml
```diff
@@ -125,26 +125,64 @@ e1:
...
+		Voice: Attack
+		RequiresCondition: deployed
+	GrantConditionOnAttack:
+		Condition: AnimateTurret
+		RevokeDelay: 10
+		ArmamentNames: deployed
+	RejectsOrders@deployment:
+		Reject: Move, EnterTransport, EnterTransports
+		RequiresCondition: deployed
+	-TakeCover: # TODO: needs to be conditional so we can disable when deployed
```
Phrohdoh: Is there an upstream ticket for this?
Would be good to create one (and link it to this PR as a use-case) if not.

MustaphaTR: I think there is but should check.

mods/ra2/rules/allied-infantry.yaml
```diff
@@ -125,26 +125,64 @@ e1:
...
+	Armament@elite-deployed:
+		Weapon: paraE
+		Name: deployed
+		RequiresCondition: rank-elite && deployed
+	GrantConditionOnDeploy:
+		DeployedCondition: deployed
+		UndeployedCondition: undeployed
+		CanDeployOnRamps: True
+		Facing: 160
+		RequiresCondition: !parachute
```
Phrohdoh: We do want to be able to give the deploy command while they are paradropping (when they hit the ground they should immediately deploy if given the command in the air).

Is this not possible technically yet?

mods/ra2/rules/allied-infantry.yaml
```diff
@@ -125,26 +125,64 @@ e1:
...
+		Name: inviso
+	AttackFrontal:
+		Voice: Attack
+		RequiresCondition: undeployed
+	AttackTurreted@deployed:
+		Armaments: deployed
+		Voice: Attack
+		RequiresCondition: deployed
+	GrantConditionOnAttack:
+		Condition: AnimateTurret
```
Phrohdoh: Nit: The casing here is inconsistent as far as I can tell.

Can we change this to `animate-turret` or `animateTurret`, whichever is more consistent throughout the codebase?

mods/ra2/rules/allied-infantry.yaml
MustaphaTR: It isn't possible yet. If i leave this out, it would make them deploy in air.

mods/ra2/rules/allied-infantry.yaml
MustaphaTR: I think this should be 128 (probably leftover from SP Marauder, which is where i copied the code from).

What is required to make the turret realign after undeploy, but not while deployed, so after deploy the turret facing won't jump. There should be an issue to make turret realign before deploy on upstream.

There is only one anim so not forcing a facing (not defining this) doesn't really help.

mods/ra2/rules/allied-infantry.yaml
MustaphaTR: https://github.com/OpenRA/OpenRA/issues/15167 Issue already mentions GIs.

mods/ra2/rules/allied-infantry.yaml
Phrohdoh: Let's create an upstream ticket and link it too, then.

mods/ra2/rules/allied-infantry.yaml
Phrohdoh: Let's link the ticket here if we can and change this to 128 then if you believe that value is more correct.

MustaphaTR:
Updated with pretty hack solutions to issues found in IRC.

mods/ra2/sequences/allied-infantry.yaml
```diff
@@ -52,6 +52,8 @@ e1:
 	Inherits: ^ArmedInfantry
 	Inherits@MC: ^MindControllable
 	Defaults: gi
+	dot: dot
+		ZOffset: -9999
```
abcdefg30: This looks like a workaround/hack. What speaks against adding the `WithDeployingInfantryBody` suggested on IRC?

abcdefg30:
I gave this a shot at #547.

PR #499, opened 2018-11-02
Change the unrestricted techlevel to provide the superweapons prereq

### LINKED ISSUE #469: Support no superweapons please
openra engine already supports this, please add it to this mod as well.

Phrohdoh:
This essentially changes the **Medium** techlevel to "no superweapons."

Closes #469.

### DIFF
--- a/mods/ra2/rules/player.yaml
+++ b/mods/ra2/rules/player.yaml
@@ -74,7 +74,7 @@ Player:
 		Id: medium
 	ProvidesTechPrerequisite@unrestricted:
 		Name: Unrestricted
-		Prerequisites: techlevel.infonly, techlevel.low, techlevel.medium, techlevel.unrestricted
+		Prerequisites: techlevel.infonly, techlevel.low, techlevel.medium, techlevel.superweapons
 		Id: unrestricted
 	GrantConditionOnPrerequisiteManager:
 	VeteranProductionIconOverlay:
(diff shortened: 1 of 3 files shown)


MustaphaTR:
I don't think tech levels were properly set up at all, so this will probably need a proper rework. We should expand this to have same tech levels as in RA mod.

Phrohdoh:
> We should expand this to have same tech levels as in RA mod.

Why match RA?

If that is what original RA2 did then great but I don't see a reason to match RA just because this game is a continuation.

That aside, I agree, the tech levels here seem arbitrary and not properly taken advantage of.

MustaphaTR:
Original RA2 had 10 tech levels, but it wasn't settable in game (tho afaik CnCNet 5 Client has it). It was mostly used for mission maps.

10 tech levels are unnecessarily too much (original RA had 12 TL, D2k had 7), what we have in RA mod is more than enough. You can even scrap the infantry only (it doesn't exits on TD and D2k mods either).

mods/ra2/rules/player.yaml
```diff
@@ -74,7 +74,7 @@ Player:
 		Id: medium
 	ProvidesTechPrerequisite@unrestricted:
```
abcdefg30: You probably want to rename those, too.

Phrohdoh: I didn't rename them because we'd have `Low`, `No Superpowers`, `Unrestricted` which feels like going from 0 to 100 _very quickly_.

I think it'll be best to close this PR and rethink/redo the tech levels as @MustaphaTR mentioned, opinions?

abcdefg30: If you want to. The current changes make sense to me, but when a major rework of the levels is coming soon anyway, it is indeed best to close this PR.

abcdefg30:
Closing as discussed.

PR #502, opened 2018-11-02
Some Value Fixes

MustaphaTR:
Commited each one seperately.

Wanted to handle these seperately from the speed fixes. So created this PR.


Phrohdoh:
These came directly from the `rules.ini`, I assume?

MustaphaTR:
Yes. ZxGanon found the issues and fixed him. But i also checked the values while committing the changes.

ZxGanon:
Yep the RA2 machinery is gonna role out soon.

Issue #151, opened 2016-02-22
Aircraft carrier


Mailaender:
See http://modenc.renegadeprojects.com/Spawns and https://youtu.be/5J2j-NUcS4E

MustaphaTR:
There is actually another issue for it. I think this can be closed as duplicate. #118

GraionDilach:
That one is about missiles, not aircraft carrier. Yes, both are Spawns-derived in RA2, but that does not directly mean the two will be related in OpenRA.

Mailaender:
There is also an upstream issue about it: https://github.com/OpenRA/OpenRA/issues/4157

Mailaender:
Just saw that @forcecore already wrote `CarrierSlave` and `CarrierMaster` traits in his private fork.

forcecore:
I have V3, too
https://www.youtube.com/watch?v=xTJNAAxUyU0 : if you watch closely, AA guns shoot down some of the missiles.

Mailaender:
V3 is #118. Can you contribute your code?

forcecore:
Indeed I would. However, it will take some time as the codes went through a major rewrite and need some stabilization.

PR #508, opened 2018-11-03
Add Aircraft Carrier

MustaphaTR:
Currently the planes are not VTOL. But otherwise it is working pretty good.

I made all the planes have `MaximumPitch: 128` so Harrier and Beagle won't go to other side of the map to land.

Closes https://github.com/OpenRA/ra2/issues/151.

### DIFF
--- a/OpenRA.Mods.RA2/Activities/EnterCarrierMaster.cs
+++ b/OpenRA.Mods.RA2/Activities/EnterCarrierMaster.cs
@@ -1,0 +1,82 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2018 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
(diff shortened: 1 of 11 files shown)


MustaphaTR:
> OpenRA usually uses child/parent instead of slave/master.

I didn't name the stuff, if you want me to rename them, i can.

> Don't we only need the carrier spawner? Why are there two different traits anyway?

The BaseSpawner/Master traits handle the base stuff which will be used by V3/Dred logic. BoolBada's code also had those stuff for Angry Mob and Slave Miner, which also used those, but we don't need them here.

If you need master/slave difference, i'm not sure about that.

> There are a couple of unnecessary comments

BoolBada was a bit over the top with the comments, yeah some may be unnecessary.

Mailaender:
This works fairly well in-game. :+1:

sorcerer86pt:
Please rebase with master

DylanC:
@MustaphaTR ^

sorcerer86pt:
Okay, why can't this commit be merged in the codebase?

GraionDilach:
Untested in this PR, but the version included in RV has an error within the code where planes above the carrier bug out when the target is destroyed and unable to enter to a stopped carrier, just fly above the entering point until the carrier is moved further and the plane can retry entering in the new place.

There are still also a good amount of coding style errors included, which are all coming from @forcecore's legacy.

For the record, I am working fairly slowly to have this codebase cleaned up for AS integration, which probably will be a better starting point for inclusion into an OpenRA repository than this current iteration is.

MustaphaTR:
I think this can be closed since @Mailaender opened his own.

Issue #411, opened 2018-01-21
Most of civilian stuff are using wrong name


MustaphaTR:
They are mostly using whatever came from `Name=` field in the rules.ini. But actual name used by RA2 is the matching `UIName=` from `ra2.csf`.

MustaphaTR:
This issue is about buildings not units.

PR #521, opened 2018-11-15
Fix civilians' name

MustaphaTR:
They were all called "Civilian" in original game. Those values are from the `Name=` field which isn't used ingame.

Closes https://github.com/OpenRA/ra2/issues/411.


MustaphaTR:
Updated.

abcdefg30:
No activity since a while, so I took the freedom to continue this at #546.

PR #524, opened 2018-11-17
Add Concrete Low Bridges

MustaphaTR:
Also adds a map that uses them (Montana DMZ).


Mailaender:
Looks a bit unpolished as the damaged variants are missing on destruction:

image

but it otherwise works fine. :heavy_check_mark:

Mailaender:
You added them manually? I still have https://github.com/Mailaender/OpenRA/tree/ra2-lobrdb lying around. Is it still useful?

MustaphaTR:
RV's DLL has its own importer command, i should've added it here too. I think i imported those map with that importer but not 100% sure.

https://github.com/MustaphaTR/Romanovs-Vengeance/blob/master/OpenRA.Mods.RA2/UtilityCommands/ImportRA2MapCommand.cs

Mailaender:
You could submit that minus the Yuri's Revenge stuff.

PR #535, opened 2019-01-09
Corrected exit positions for Airforce Command Headquarters

Nihilianth:
This corrects the positions of Harriers and Black Eagles when grounded mentioned in #495 

Before:
2019-01-09 18_04_51-openra
After:
2019-01-09 18_10_53-openra


Mailaender:
Thanks!

PR #546, opened 2019-03-02
Fix civilian names

### LINKED ISSUE #411: Most of civilian stuff are using wrong name

abcdefg30:
Supersedes #521.

`Name: Civilian` is needed as some civilians don't have a name assigned to them and would then get an empty tooltip with the visibility cheat enabled.

Closes #411.


MustaphaTR:
Just to note that one of the civilians had wrong name. I don't remember which tho.

PR #569, opened 2019-06-22
Fix Gap Generator footprint

### LINKED ISSUE #568: Gap Generator footprint is wrong.
It has a bib cell. RA2 GapGen was 1x1.

MustaphaTR:
Fixes #568.


GraionDilach:
Pretty sure that's because this mod lacks depthbuffer usage.

abcdefg30:
I guess we'll have to revisit a lot of the offsets anyway when adding depth.

abcdefg30:
The build does some weird things, but this is not an issue of this PR.

PR #573, opened 2019-07-08
Update engine version to release-20190314.

### LINKED ISSUE #572: Use Latest Released OpenRA Engine
Lots of changes between releases and RA2 hasn't kept up-to-date.

I had an honest go at a patch, but it still needs work before the game will run.

[release-20190314.patch.txt](https://github.com/OpenRA/ra2/files/3365299/release-20190314.patch.txt)

MustaphaTR:
Closes #572.

### DIFF
--- a/OpenRA.Mods.RA2/Traits/Render/WithExitOverlay.cs
+++ b/OpenRA.Mods.RA2/Traits/Render/WithExitOverlay.cs
@@ -19,7 +19,7 @@ using OpenRA.Traits;
 namespace OpenRA.Mods.RA2.Traits
 {
 	[Desc("Renders an animation when when the actor is leaving from a production building.")]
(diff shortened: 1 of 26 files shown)


Mailaender:
Can you update `.travis.yml` to match https://github.com/OpenRA/OpenRA/commit/15e88d9e58a4da5c6b9aa5389561e3e18f61bce2#diff-354f30a63fb0907d4ad57269548329e3R6 as I hope that fixes the build.

OpenRA.Mods.RA2/Traits/Render/WithExitOverlay.cs
```diff
@@ -84,8 +71,16 @@ void INotifyProduction.UnitProduced(Actor self, Actor other, CPos exit)
 
 		void ITick.Tick(Actor self)
 		{
+            if (IsTraitDisabled)
```
Mailaender: Spaces instead of tabs here and below.

Issue #564, opened 2019-06-09
Tech Outpost shadow issues.


GraionDilach:
Tech Outpost have 6 frames on Temperate and Urban, but 8 frames on Snow (due to WW forgetting to include the rubble frames). This throws ShadowStart off.

indivisibleman:
I think this is fixed by #578

GraionDilach:
Eyup it was.

PR #578, opened 2019-08-18
Added rubble for Airport, Hospital, and Oil Derrick

### LINKED ISSUE #555: Tech Structures are missing rubbles.

indivisibleman:
Rubble art added for Airport, Hospital, and Oil Derrick Tech Structures to fix #555. Rubble for the Tech Outpost should be dealt with in #564 due to different numbers of animation frames for each tile set.

### DIFF
--- a/mods/ra2/rules/palettes.yaml
+++ b/mods/ra2/rules/palettes.yaml
@@ -53,6 +53,10 @@
 		Tileset: URBAN
 		Filename: liburb.pal
 		ShadowIndex: 1
+	PaletteFromFile@outpost-rubble:
+		Name: outpost-rubble
+		Filename: isosno.pal
+		ShadowIndex: 1
 	PaletteFromFile@playerurb:
 		Name: player
 		Tileset: URBAN
--- a/mods/ra2/rules/tech-structures.yaml
+++ b/mods/ra2/rules/tech-structures.yaml
@@ -25,6 +25,17 @@ caoild:
 		Offset: 1000,0,0
 	GivesCashOnCapture:
 		Amount: 1000
+	SpawnActorOnDeath:
+		Actor: caoild.rubble
+
+caoild.rubble:
+	Inherits: ^Rubble
+	Inherits@shape: ^2x2Shape
+	Building:
+		Dimensions: 2,2
+		Footprint: == ==
+	RenderSprites:
+		Image: caoild
 
 caairp:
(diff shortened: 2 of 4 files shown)


GraionDilach:
Tech Outpost should use the snow rubble in all theaters. That is doable and that's what RA2 modders tend to do.

For the issues with the standard artwork, the best option probably is to Combine: the sequences together.

indivisibleman:
I don't know how to combine the sequences yet, is there an example somewhere? Should that be done for this pull request or a separate one?

MustaphaTR:
There should be a non-snow Outpost rubble i made somewhere on PPM. Tho i think it was just PNG image posted there. I'll try to get the SHP out from my mod tomorrow.

Flame Tank muzzle animation in TD should br using Combine:.

GraionDilach:
Doublechecked the source code and one cannot combine tileset-dependent artwork. So combining is not an option. In AS, I worked around this via splitting the shadows into WithIdleOverlays.

indivisibleman:
How would you get separate shadows for each tileset using WithIdleOverlays?

GraionDilach:
Add this to rules:
```
caoutp:
	WithIdleOverlay@tower-shadow:
		Sequence: idle-tower-shadow
```

and modify sequences:
```
caoutp:
	idle-tower:
		ShadowStart: -1 # to disable the external shadow frames
	damaged-idle-tower:
		Start: 1
		ShadowStart: -1
	idle-tower-shadow:
		UseTilesetCode: false # to make sure that shadows are taken from one file and does not vary with theater variants
		Start: 4
	damaged-idle-tower-shadow:
		UseTilesetCode: false
		Start: 5
```

indivisibleman:
I've added the code you suggested to sort out shadows, this fixes #564. I've also added the rubble for the Tech Outpost, but it is rendered in the palette for the current map, is there a way to have it use the isosno.pal?

GraionDilach:
Somewhere within rules/palettes.yaml:
```
	PaletteFromFile@outpost-rubble:
		Name: outpost-rubble
		Filename: isosno.pal
		ShadowIndex: 1
```
and then update the rules:

```
caoutp.rubble:
	RenderSprites:
		Image: caoutp
		Palette: outpost-rubble
```

indivisibleman:
Thanks. It now uses the correct palette.

PR #595, opened 2019-12-21
UI: Power meter display

torleif:
This enhancement will display the power usage in the side navbar, similar to how the original RA2 did. It features a nice animation effect, scales to the size of the sidebar, and a percentage fallback when too much power is produced. 

The parameters can be modified through yml. 

power-meter

### DIFF
--- a/OpenRA.Mods.RA2/Widgets/PowerMeterWidget.cs
+++ b/OpenRA.Mods.RA2/Widgets/PowerMeterWidget.cs
@@ -1,0 +1,250 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2020 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Mods.Common.Widgets;
+using OpenRA.Widgets;
+using OpenRA.Graphics;
+using System;
+
+namespace OpenRA.Mods.RA2.Widgets.Logic
+{
+
+	public class PowerMeterWidget : Widget
+	{
+		Widget sidebarProduction;
+
+		int lastMeterCheck;
+
+		int barheight = 0;
+
+		bool bypassanimation = false;
+
+		int warningflash = 0;
+
+		int lasttotalpowerdisplay;
+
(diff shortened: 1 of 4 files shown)


OpenRA.Mods.RA2/Widgets/PowerMeterWidget.cs
```diff
@@ -0,0 +1,250 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2018 The OpenRA Developers (see AUTHORS)
```
Mailaender: 2019, but we can update that with all the other files in 2 days anyway.

OpenRA.Mods.RA2/Widgets/PowerMeterWidget.cs
```diff
@@ -0,0 +1,250 @@
...
+using System;
+
+namespace OpenRA.Mods.RA2.Widgets.Logic
+{
+
+	public class PowerMeterWidget : Widget
+	{
+		Widget sidebar_production;
+
+		[Sync] int lastMeterCheck;
```
Mailaender: Why does the UI need to sync between players? I assume this does nothing as the `ISync` interface isn't implemented.

OpenRA.Mods.RA2/Widgets/PowerMeterWidget.cs
```diff
@@ -0,0 +1,250 @@
...
+		public readonly string FlashPowerImage = "";
+
+		[Desc("The collection of images to get the meter images from")]
+		[FieldLoader.Require]
+		public readonly string ImageCollection = "";
+
+		[ObjectCreator.UseCtor]
+		public PowerMeterWidget(World world, WorldRenderer worldRenderer)
+		{
+			myworld = world;
```
Mailaender: Style nit: we usually write `this.world = world`.

OpenRA.Mods.RA2/Widgets/PowerMeterWidget.cs
```diff
@@ -0,0 +1,250 @@
...
+using OpenRA.Widgets;
+using OpenRA.Graphics;
+using System;
+
+namespace OpenRA.Mods.RA2.Widgets.Logic
+{
+
+	public class PowerMeterWidget : Widget
+	{
+		Widget sidebar_production;
```
Mailaender: Can we use pascalCase here?

Mailaender:
This looks and plays fantastic. I have only minor reservations. Also, the commit message could have a cleanup.

torleif:
@Mailaender thanks for the review! i've made the recommended changes

PR #618, opened 2020-01-31
Make Ore Purifier work.

### LINKED ISSUE #289: Ore purifier logic
https://github.com/OpenRA/OpenRA/issues/11833

MustaphaTR:
The logic has been in the engine for a while now, but doesn't seem like it was coded to be used by the purifier.

Fixes  #289

### DIFF
--- a/mods/ra2/rules/allied-structures.yaml
+++ b/mods/ra2/rules/allied-structures.yaml
@@ -811,6 +811,8 @@ gaorep:
 	WithIdleOverlay@glow:
 		Sequence: idle-glow
 		RequiresCondition: !build-incomplete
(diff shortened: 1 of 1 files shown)


mods/ra2/rules/allied-structures.yaml
```diff
@@ -811,6 +811,8 @@ gaorep:
 	WithIdleOverlay@glow:
 		Sequence: idle-glow
 		RequiresCondition: !build-incomplete
+	ResourcePurifier:
+		Modifier: 25
```
sorcerer86pt: This means that with the purifier on, we get +25% of all income? is that it?

PR #621, opened 2020-02-01
Fix Frame order for tib02 and tib14.

### LINKED ISSUE #561: tib02 and tib14 have wrong frame orders
They start with the last frame. Original RA2 bug though but would be nice as a polish.

MustaphaTR:
Fixes  #561.


sorcerer86pt:
Can you add in the commit message the why of this change. Aside from that ok

Issue #583, opened 2019-09-18
Investigate duplicate `modcontent` mod


Phrohdoh:
Can the engine one be removed from the source tree?

PR #625, opened 2020-02-15
Removed the redundant mod content mod

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/583.


MustaphaTR:
`PACKAGING_COPY_ENGINE_FILES="./mods/modcontent"` needs to be defined in `mod.config` or it won't work in installer versions.

Mailaender:
Added.

Issue #182, opened 2016-04-03
War factories don't open roofs for aircraft


Mailaender:
Nighthawk and Kirov are the examples.

cookgreen:
Here is my idea: When War factory build aircraft finished, War Factory will be replaced by the War Factory which open roof, and set a time like 2 or 3 secs to trigger the War Factory become the original state after it open roof

GraionDilach:
https://github.com/OpenRA/OpenRA/pull/15303 providing a condition to trigger the open-roof animations with a duration synced with the aircraft flying up sounds like the simplest solution (it always flies up afterall, so jammed exits can't be an issue).

PR #626, opened 2020-02-15
Fixed the roofs not opening when building aircraft units

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/182.

### DIFF
--- a/OpenRA.Mods.RA2/Traits/Render/WithExitOverlay.cs
+++ b/OpenRA.Mods.RA2/Traits/Render/WithExitOverlay.cs
@@ -1,6 +1,6 @@
 #region Copyright & License Information
 /*
- * Copyright 2007-2019 The OpenRA Developers (see AUTHORS)
+ * Copyright 2007-2020 The OpenRA Developers (see AUTHORS)
  * This file is part of OpenRA, which is free software. It is made
  * available to you under the terms of the GNU General Public License
  * as published by the Free Software Foundation, either version 3 of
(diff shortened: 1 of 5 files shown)


OpenRA.Mods.RA2/Traits/Render/WithExitOverlay.cs
```diff
@@ -62,24 +64,16 @@ void INotifyDamageStateChanged.DamageStateChanged(Actor self, AttackInfo e)
 			overlay.ReplaceAnim(RenderSprites.NormalizeSequence(overlay, e.DamageState, overlay.CurrentSequence.Name));
 		}
 
-		void INotifyProduction.UnitProduced(Actor self, Actor other, CPos exit)
+		public void UnitProduced(Actor self, Actor other, CPos exit)
```
abcdefg30: What is this change for?

Mailaender: That was an accident.

mods/ra2/rules/allied-structures.yaml
```diff
@@ -397,12 +397,28 @@ gaweap:
...
+	GrantConditionOnProduction@helicopters:
+		Actors: shad
+		Condition: roof-open
+		Duration: 50
+		ShowSelectionBar: false
+	GrantConditionOnProduction@zeppelins:
+		Actors: zep
+		Condition: roof-open
+		Duration: 100
+		ShowSelectionBar: false
```
MustaphaTR: This seems to be not working but it is an upstream issue.

Mailaender: I see a pink selection bar during the roof opening.

MustaphaTR: Yes, that's what the value is meant to remove.

Mailaender: Fixed https://github.com/OpenRA/OpenRA/pull/17704.

MustaphaTR:
I guess this can be merged regardless of that bug so doing it.

Issue #131, opened 2016-02-12
Polish MCV packing logic


Fincer:
In the original Red Alert 2, you can pack and move Construction Yard by simply giving a move order to the building itself:

![mcvunpack-ra2-cropped-comp](https://cloud.githubusercontent.com/assets/3705231/13010360/26965120-d1aa-11e5-915c-1f2c5fbfc382.gif)

In RA2 mod (OpenRA), you need to tell Construction Yard to pack itself and after that you can move it:

![mcvunpack-openra2-cropped-comp](https://cloud.githubusercontent.com/assets/3705231/13010364/2bba3bbc-d1aa-11e5-8e81-26ddfb17cf7a.gif)

Please pay attention to cursor icons. In RA2, the cursor indicates I can move the building. Not so in OpenRA.

OpenRA/ra2 approach is not so intuitive than seen in the original game. Additionally, I feel it's unnecessarily compilated and in a hectic battle there should be a way to tell MCV just move quickly away, not telling it first to pack itself while a player need to pay attention to the battle events at the same time. Now it feels there is an unnecessary extra command for no good reason.

Mailaender:
That is why the game doesn't have an undeploy cursor.

penev92:
One accidentally cancelled Nuke silo, coming up!

GraionDilach:
Actually, this has pros and cons. A consequence of this in OpenRA is that you can finally set rallypoints for mobile war factories - in RA2 trying to set those resulted with the actual factory undeploying.

MustaphaTR:
I think it should be settable under `Transforms:` so we can use it for CYards but modders may disable it for MWFs.

scshunt:
It would work, I think, if you required a force-move (ALT) to do it. That would prevent accidental deploys and still allow rally points.

I think the big trouble here is the way that the conyard decides what moves are allowed. The conyard doesn't know about the movement types of the unit its becoming, and there's a lot of logic that it's not equipped to handle. We need a trait to do those calculations, but we can't easily just put one onto the conyard (what trait would that be, even? What if you have a flying MCV that uses Aircraft rather than Mobile?).

Once such a trait was written, we would then need to communicate the movement order to the newly-undeployed MCV. That would be the easier part, I think, but would still require changes to the transformation process.

GraionDilach:
FWIW this issue should probably be moved to upstream because TS had this as well (in TS, rally points were ordered with Alt-click, which is why the two could be combined).

[...]

PR #627, opened 2020-02-15
Added force move undeploy to the construction yards

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/131.

### DIFF
--- a/mods/ra2/rules/allied-structures.yaml
+++ b/mods/ra2/rules/allied-structures.yaml
@@ -30,6 +30,21 @@ gacnst:
 		Facing: 96
 		DeployCursor: undeploy
 		RequiresCondition: factundeploy && !build-incomplete
+	TransformsIntoMobile:
+		RequiresCondition: factundeploy
+		Locomotor: heavytracked
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoRepairable:
+		RequiresCondition: factundeploy
+		RepairActors: gadept, nadept, caoutp
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoPassenger:
+		RequiresCondition: factundeploy
+		CargoType: Vehicle
+		Voice: Move
+		RequiresForceMove: true
 	GrantConditionOnPrerequisite@GLOBALFACTUNDEPLOY:
 		Condition: factundeploy
 		Prerequisites: global-factundeploy
--- a/mods/ra2/rules/soviet-structures.yaml
+++ b/mods/ra2/rules/soviet-structures.yaml
@@ -32,6 +32,21 @@ nacnst:
 		Facing: 96
 		DeployCursor: undeploy
 		RequiresCondition: factundeploy && !build-incomplete
+	TransformsIntoMobile:
+		RequiresCondition: factundeploy
+		Locomotor: heavytracked
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoRepairable:
+		RequiresCondition: factundeploy
+		RepairActors: nadept, gadept, caoutp
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoPassenger:
+		RequiresCondition: factundeploy
+		CargoType: Vehicle
+		Voice: Move
+		RequiresForceMove: true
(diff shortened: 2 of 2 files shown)


mods/ra2/rules/allied-structures.yaml
```diff
@@ -30,6 +30,21 @@ gacnst:
...
 		DeployCursor: undeploy
 		RequiresCondition: factundeploy && !build-incomplete
+	TransformsIntoMobile:
+		RequiresCondition: factundeploy
+		Locomotor: tracked
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoRepairable:
+		RequiresCondition: factundeploy
+		RepairActors: gadept
```
MustaphaTR: These should list all 3 (caoutp too). Faction's one.

mods/ra2/rules/soviet-structures.yaml
```diff
@@ -32,6 +32,21 @@ nacnst:
...
 		DeployCursor: undeploy
 		RequiresCondition: factundeploy && !build-incomplete
+	TransformsIntoMobile:
+		RequiresCondition: factundeploy
+		Locomotor: tracked
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoRepairable:
+		RequiresCondition: factundeploy
+		RepairActors: nadept
```
MustaphaTR: Same here.

mods/ra2/rules/soviet-structures.yaml
```diff
@@ -32,6 +32,21 @@ nacnst:
...
 		DeployCursor: undeploy
 		RequiresCondition: factundeploy && !build-incomplete
+	TransformsIntoMobile:
+		RequiresCondition: factundeploy
+		Locomotor: tracked
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoRepairable:
+		RequiresCondition: factundeploy
+		RepairActors: nadept, caoutp, gadept
```
MustaphaTR: `gadept, nadept, caoutp` please. Same applies allied one.

Mailaender: It is exactly the same just in a different order.

MustaphaTR: Yes, for consistency. I'm sure other places (Repairable in defaults) use that order.

mods/ra2/rules/soviet-structures.yaml
```diff
@@ -32,6 +32,21 @@ nacnst:
 		Facing: 96
 		DeployCursor: undeploy
 		RequiresCondition: factundeploy && !build-incomplete
+	TransformsIntoMobile:
+		RequiresCondition: factundeploy
+		Locomotor: tracked
```
MustaphaTR: MCVs have `heavytracked` locomotor.

mods/ra2/rules/soviet-structures.yaml
```diff
@@ -32,6 +32,21 @@ nacnst:
...
+		Voice: Move
+		RequiresForceMove: true
+	TransformsIntoRepairable:
+		RequiresCondition: factundeploy
+		RepairActors: nadept, gadept, caoutp
```
MustaphaTR: This is till wrong.

Mailaender: `nadept` is the soviet counter-part. Why is this wrong for the soviet MCV?

MustaphaTR: All units can get repaired from all Service depots, i don't see any reason why Soviet Service Depot should be first for Soviet MCV.

Mailaender: Because that is most likely in a safe place at the mother base.

MustaphaTR: I think changing the order would be an unnecessary inconsistency in the code. The order changes nothing ingame.

Mailaender: Why are you making me change this all the time if it doesn't do anything? I think my way of setup is more consistent with the faction specific buildings.

Issue #630, opened 2020-02-15
Cursor does not have a sequence `attackmove-blocked`


pgwipeout:
OpenRA engine version playtest-20200118
Red Alert 2 mod version {DEV_VERSION}
on map 3a94af24afc3d43b59eecdff309bbe44c5047644 (Pitch Point by Westwood Studios).
Date: 2020-02-15 23:01:25Z
Operating System: Windows (Microsoft Windows NT 6.2.9200.0)
Runtime Version: .NET CLR 4.0.30319.42000
Exception of type `System.InvalidOperationException`: Cursor does not have a sequence `attackmove-blocked`
   at OpenRA.Graphics.CursorProvider.GetCursorSequence(String cursor)
   at OpenRA.Graphics.SoftwareCursor.Render(Renderer renderer)
   at OpenRA.Game.RenderTick()
   at OpenRA.Game.Loop()
   at OpenRA.Game.Run()
   at OpenRA.Game.InitializeAndRun(String[] args)
   at OpenRA.Program.Main(String[] args)

PR #631, opened 2020-02-16
Added the attackmove-blocked cursor

Mailaender:
Closed https://github.com/OpenRA/ra2/issues/630.

### DIFF
--- a/mods/ra2/cursors.yaml
+++ b/mods/ra2/cursors.yaml
@@ -292,3 +292,8 @@ Cursors:
 			Length: 9
 		assaultmove-minimap:
 			Start: 9
+		assaultmove-blocked: # TODO
+			Start: 0
+			Length: 9
+		assaultmove-blocked-minimap: # TODO
+			Start: 9
--- a/mods/ra2/cursors.yaml
+++ b/mods/ra2/cursors.yaml
@@ -61,6 +61,8 @@ Cursors:
 		attackmove:
 			Start: 404
 			Length: 9
+		attackmove-blocked:
+			Start: 384
 		attack-minimap:
 			Start: 63
 			Length: 5


MustaphaTR:
What about `assaultmove-blocked`?

Mailaender:
Added.

mods/ra2/cursors.yaml
```diff
@@ -290,3 +292,6 @@ Cursors:
 			Length: 9
 		assaultmove-minimap:
 			Start: 9
+	mouse.shp: mouse
```
abcdefg30: Why did you create a new section for this but didn't add it to the first `mouse.shp` section?

Mailaender: To keep the cursors next to each other.

abcdefg30: Can you please move it into the first section? On IRC we weren't quite sure if (ab)using the yaml merger in this way doesn't have any side effects. Keeping them separated will also be a potential bug source if anyone ever was to change the first section (not expecting it to be defined a second time).

Mailaender: It would be pretty terrible if the yaml merger couldn't resolve this. There are unit tests in place https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Test/OpenRA.Game/MiniYamlTest.cs to check for problems.

pgwipeout:
I just triggered this crash again.
Any word on getting this fix merged?

Mailaender:
Updated this to also include the minimap variants.

sorcerer86pt:
@pgwipeout Hi, can you describe what actions did you do to trigger this?

pgwipeout:
@sorcerer86pt I use the attack move hotkey a lot.
Usually it triggers if I hit the hotkey while over fog of war, I imagine it's because the cursor is over terrain that is impassible. But due to it being over fog, I can't tell you exactly where or why.
Also the crash is instantaneous, so I don't get a chance to observe the screen at the time of death.

sorcerer86pt:
Base Profile Screenshot 2020 03 19 - 13 39 36 86
Base Profile Screenshot 2020 03 19 - 13 39 57 47
Base Profile Screenshot 2020 03 19 - 13 40 04 82

Tested like @pgwipeout  mentioned, using an island map, and using GIs. Tried to launch attack move on not passable terrain to them, due to passing through water. And tried to do the same using the minimap. No crash. Also the cursor changed accordingly.

Issue #576, opened 2019-08-03
Low Power voice notifications are way too frequent.


ShikiSuen:
In original RA2 and YR, when your base goes insufficient power, there is only one voice notification of "low power" until your power becomes sufficient.

In current RA2 mod for OpenRA, every 5 seconds comes one voice notification of "low power". This is not friendly to those people who have anxiety disorder.

PR #633, opened 2020-02-16
Fixed the low power notification spam

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/576.


ShikiSuen:
Finally. Thanks for the fix.

Issue #557, opened 2019-04-14
[Enhancement] Apply RA2 bubble effect in chronoshift


sorcerer86pt:
It's just a visual enhancement. At the moment, chronoshift just teleport the units in the area. In RA2 it had that double bubble animation (creating a energy bubble in the units, then another at the destination spot, while teleporting the units).

![chrono shift](https://vignette.wikia.nocookie.net/cnc/images/a/ae/CNCRA2_Chronosphere_Demonstration.png/revision/latest?cb=20160109031002)

PR #650, opened 2020-03-08
Added ChronoshiftableWithSpriteEffect

Mailaender:
Starts tackling https://github.com/OpenRA/ra2/issues/557.


sorcerer86pt:
Just remember that Chrono legionnaires in OG Ra2 can be chronosphered successfully (exception to major support powers killing infantary on use)

Mailaender:
You are right that the bubble is still missing. That might need an engine update.

Mailaender:
Okay. Done.

sorcerer86pt:
ok, finished testing it. It looks good. the z-offset make the effect going directly down on the vehicles.

PR #651, opened 2020-03-08
Added SpawnSurvivors

Mailaender:
However, I removed the redundancy towards `SpawnActorsOnSell`.

### DIFF
--- a/OpenRA.Mods.RA2/Traits/SpawnSurvivors.cs
+++ b/OpenRA.Mods.RA2/Traits/SpawnSurvivors.cs
@@ -1,0 +1,74 @@
+﻿#region Copyright & License Information
+/*
+ * Copyright 2007-2020 The OpenRA Developers (see AUTHORS)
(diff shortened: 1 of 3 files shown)


MustaphaTR:
Current yaml setup makes it so the game checks for the actual faction of the building while selling but use the matching faction with the building on destruction. Altho this shouldn't matter in a normal game unless you get an MCV of other faction from a crate, it is an unnecessary inconsistency.

sorcerer86pt:
okay, both destroying and selling spawn survivors.
For example soviet mcv sell/destroyed spawns 1 eng, 7 concripts, while allied spawns 1 eng, 4 GIs

Mailaender:
Conscripts are cheap.

sorcerer86pt:
I know, and it's perfectly okay. I was saying that because it's a
confirmation that what is described in the yaml file is being successfuly
spawned. And since conyard is the one with engineer and basic infantry, I
used it as example

PR #652, opened 2020-03-08
Setup Chronosphere and Iron Curtain animations

Mailaender:
Depends on https://github.com/OpenRA/OpenRA/pull/17787.


sorcerer86pt:
rebase this please

Mailaender:
No point in that really as https://github.com/OpenRA/OpenRA/pull/17787 isn't even merged.

sorcerer86pt:
HI, the https://github.com/OpenRA/OpenRA/pull/17787 has been merged. I think you can rebase this and retest this

Mailaender:
No, it is still https://github.com/OpenRA/ra2/blob/0bd1e2892b513aabf5a94977e2c850adb848b32c/mod.config#L7

Mailaender:
Rebased

MustaphaTR:
Needs rebase again.

Mailaender:
Had to rebase again due to https://github.com/OpenRA/ra2/pull/806 going in first.

OpenRA.Mods.RA2/Traits/Render/WithSupportPowerChargedOverlay.cs
```diff
@@ -0,0 +1,85 @@
...
+#endregion
+
+using OpenRA.Graphics;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Mods.Common.Traits.Render;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA2.Traits.Render
+{
+	[Desc("Displays an overlay when `SupportPower` is fully charged.")]
```
penev92, line 19:
```suggestion
	[Desc("Displays an overlay when `" + nameof(SupportPower) + "` is fully charged.")]
```

OpenRA.Mods.RA2/Traits/Render/WithSupportPowerChargedOverlay.cs
```diff
@@ -0,0 +1,85 @@
...
+	public class WithSupportPowerChargedOverlayInfo : ConditionalTraitInfo, Requires<RenderSpritesInfo>, Requires<BodyOrientationInfo>
```
penev92, line 20: This should also have `Requires<SupportPowerInfo>`.

OpenRA.Mods.RA2/Traits/Render/WithSupportPowerChargedOverlay.cs
```diff
@@ -0,0 +1,85 @@
...
+		public readonly string LoopSequence = "loop";
+
+		[SequenceReference]
+		[Desc("Sequence name to use")]
+		public readonly string EndSequence = "end";
+
+		[Desc("Position relative to body")]
+		public readonly WVec Offset = WVec.Zero;
+
+		[PaletteReference("IsPlayerPalette")]
```
penev92, line 37:
```suggestion
		[PaletteReference(nameof(IsPlayerPalette))]
```

OpenRA.Mods.RA2/Traits/Render/WithSupportPowerChargedOverlay.cs
Mailaender: No, it is loosely coupled with interfaces.

penev92: It won't work at all on an actor that doesn't have a `SupportPower` trait.

Mailaender: See [WithSupportPowerActivationOverlay](https://github.com/OpenRA/OpenRA/blob/a261abbc3f6c5dbf80b0e5132a0071efdd099bc5/OpenRA.Mods.Common/Traits/Render/WithSupportPowerActivationOverlay.cs#L18). This is only to be used for trait queries that otherwise crash the game. The first lines already mentions everything you need to know for the documentation.

penev92:
Re using `Requires<SupportPowerInfo>` - I see that the argument is that we only use that for cases that would crash the game and not cases where otherwise unwanted behaviour would be introduced. That is an upstream convention and while I think it should change I won't hold it against this PR.

Mailaender:
I think the main purpose is to modify the initialization order, which is not required in case of these rendering traits which are connected via interfaces.

penev92:
Here are some polish points to consider:
 - The Weather Control Device's charged animation should freeze when low on power. IMO it would be best to remove the lightnings, not freeze them, but currently nothing happens, which seems pretty weird.
 - The Nuclear Missile Silo's charged and firing animations play on top of the default one, meaning it is clearly visible underneath! Also the missile sprite has lower ZOffset than the Silo's firing sprite, so the missile comes up from behind the structure when firing.
  - The Chronosphere's animations feel like they play too fast (I think we've mentioned this about other building animations as well some time?), but I'll need to compare to the original. The charged animation should also freeze on low power though.
  - The Iron Curtain's animations play too fast. They should also freeze on low power.

If the animation speed is a general mod thing not specific to these structures it can be addressed in a separate PR.

mods/ra2/mod.yaml
```diff
@@ -88,7 +89,6 @@ Sequences:
 	ra2|sequences/civilian-props.yaml
 	ra2|sequences/civilian-structures.yaml
 	ra2|sequences/defaults.yaml
-	ra2|sequences/misc.yaml
```
penev92, line 91: With the move of the resource sequences inside the file, is moving the file reference still needed?

penev92:
Considering how we currently have the new `WithSupportPowerChargedOverlay` trait play on top of the regular body/animation/overlay trait(s), wouldn't it be better to instead create a trait that grants a condition on support power charged and have that condition enable an overlay (disabling other rendering traits as needed)? This sounds a bit more generic and reusable.

mods/ra2/mod.yaml
Mailaender: The whole issue seemed to be resolved upstream since 2 years ago.

Mailaender:
From my experience, rendering traits work better with interfaces than loosely coupled with conditions.

PR #653, opened 2020-03-10
Set the chronosphere target palette

Mailaender:
to the same we use for building placement.


sorcerer86pt:
LGTM

Base Profile Screenshot 2020 03 15 - 19 57 48 35
Base Profile Screenshot 2020 03 15 - 19 57 52 69
Base Profile Screenshot 2020 03 15 - 19 57 57 01

Issue #43, opened 2016-01-16
Desolitor shot have not original weapon rendering?


yaaaaa:
![default](https://cloud.githubusercontent.com/assets/11130958/12372204/13cffb58-bc7a-11e5-9585-6ec74c5021f4.png)

last time I used Desolitors in Phrohdoh's repo, and it'll be fine.

GraionDilach:
Line (contrail, laser) widths have been readjusted since in upstream and the fallback default ended up being scaled with cellsize, which due to RA2 having huge tiles compared to say RA1 ended up wider as well.

Also, Desos never had their RA2 effect, because that one is a sinus wave and not a line.

yaaaaa:
maybe I have something confused. probably baffled me that before this wave has a different color

MustaphaTR:
That would be better if we change color of it to green as a start. Purple looks too unreleated.

Phrohdoh:
Upstream does not have paletted line rendering yet so we cannot remap the line drawn, so there is no point changing it yet.

@abcdefg30 what is the bug here?

GraionDilach:
The Desolator's beam wasn't remappable, but green - https://github.com/Phrohdoh/oramod-ra2/blob/master/extra-files/rules.ini#L752. Both the width and colors are bug.

Phrohdoh:
[Generate a sine wave](http://math.stackexchange.com/questions/196207/drawing-sine-and-cosine-waves) and pass the world-space points to [WorldRenderer.ScreenPosition](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Graphics/WorldRenderer.cs#L206) then those `float2`s to [the RgbaColorRenderer](https://github.com/OpenRA/OpenRA/blob/bleed/OpenRA.Game/Graphics/RgbaColorRenderer.cs).

Phrohdoh:
[WIP](https://github.com/Phrohdoh/OpenRA/tree/sine-beam):
![deso-test-03](https://cloud.githubusercontent.com/assets/4861023/16424083/48a7ae72-3d24-11e6-8847-25dfbd0e7e77.gif)

forcecore:
With my Railgun PR in bleed, I think I have a good chance of getting the sine beam into upstream, too. The code won't be too different.

Nihilianth:
[...]

PR #656, opened 2020-03-14
Added the radiation beam renderable

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/43.

### DIFF
--- a/mods/ra2/weapons/zaps.yaml
+++ b/mods/ra2/weapons/zaps.yaml
@@ -173,13 +173,13 @@ RadBeamWeapon:
 	Range: 6c0
 	ReloadDelay: 100
 	Report: idesat1a.wav
-	Projectile: LaserZap
-		Width: 20
-		Duration: 7
-		Color: B900FF00
+	Projectile: RadBeam
+		Amplitude: 476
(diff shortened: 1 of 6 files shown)


Mailaender:
I think it was done to avoid rounding errors. https://github.com/AttacqueSuperior/ValiantShades/commit/f51fc3f57e3271027f363d052dbd6787f3edac53 I accidentally copied and pasted it.

mods/ra2/weapons/zaps.yaml
```diff
@@ -170,13 +170,13 @@ SuperComet:
 
 RadBeamWeapon:
 	ValidTargets: Ground
+	InvalidTargets: Structure, Wall, ImmuneToRadiation
```
MustaphaTR: Units that were immune to radiation were still targetable by Desolator in the original game, so move this to warhead.

MustaphaTR: Also Harvesters and Other Dedolators should be immune to rad too.

Mailaender: This will probably cause problems with the AI, but okay.

Issue #86, opened 2016-01-23
Chronosphere don't teleport landed helicopters and planes


yaaaaa:
In original RA2 game, it do this too.

Mailaender:
See https://github.com/OpenRA/OpenRA/issues/17027

PR #658, opened 2020-03-14
Allowed aircraft on the ground to be chronoshifted

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/86


sorcerer86pt:
also add to this chrono legionnaires. They can be chrono shifted in OG ra2

Mailaender:
Okay done.

Issue #623, opened 2020-02-13
attack dogs are at the moment useless


sorcerer86pt:
At the moment attack dogs for both factions doesn't do anything:

- Can't attack
- Can't detect spys

only thing they can do is explore terrain.

MustaphaTR:
It looks like ^Infantry is missing EdibleByLeap: but even after adding that and editing the weapon a bit to match upstream RA, i still couldn't get it to actually kill the units. They now jump but not deal any damage.

sorcerer86pt:
So now they jump... For petting? 😋🥰👌

A terça, 18/02/2020, 4:19 da tarde, Mustafa Alperen Seki <
notifications@github.com> escreveu:

Issue #402, opened 2017-10-22
Attack dogs get veteran or elite


Icenowy:
In original RA2 attack dogs cannot get promoted, however currently in this mod they can.

abcdefg30:
Sure, go ahead. (Do note that I might not have time to look at your PR right away.)

PR #661, opened 2020-03-15
 Fixed dogs not dealing damage and removed veterancy

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/623.
Closes https://github.com/OpenRA/ra2/issues/402.

### DIFF
--- a/mods/ra2/rules/allied-infantry.yaml
+++ b/mods/ra2/rules/allied-infantry.yaml
@@ -60,6 +60,19 @@ dog:
 		HP: 100
 	Mobile:
 		Speed: 120
+	-GainsExperience:
+	-FirepowerMultiplier@VETERAN:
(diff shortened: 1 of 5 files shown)


MustaphaTR:
I was about to comment,

> Can you move the `Inherits: ^GainsExperience` to each actor that gains XP specifically instead of removing all these here. There are more than dogs that shouldn't be able to gain XP.

for the veterancy changes. But you merged before i could. It looks weird with so many removals for one thing.

Mailaender:
Sorry for being too quick. I also thought about it. Our setup is a bit weird that dogs are infantry after all. This indeed might need a cleanup.

MustaphaTR:
`^GainsExperience` is being defined on each actor, instead of defaults, on upstream for a good while now. That's also the case in RV.

PR #662, opened 2020-03-15
Set up Muzzle Animations

MustaphaTR:
I implemented this in RV and wanted to port it here too.


Mailaender:
> ./utility.sh --check-sequence-sprites
Tileset: Temperate
Tileset: Snow
Tileset: Urban

OK

Mailaender:
Thanks. Much appreciated. That looks like a lot of effort gone into this.

PR #668, opened 2020-03-16
Added smoke to the nuclear missile silo

Mailaender:


Mailaender:
Conflicts with https://github.com/OpenRA/ra2/pull/652.

Issue #567, opened 2019-06-16
Actors who are freed from mindcontrol after their owner have lost the game belong to the Neutral player.


GraionDilach:
In RA2, these actors ended up owned by Creeps instead.

GraionDilach:
Related: https://github.com/AttacqueSuperior/Engine/commit/d6e2819a0ce4bcfa5f57fb1602c5bba1a8b9ffc7

PR #669, opened 2020-03-16
Improved Mind Controller

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/567.


sorcerer86pt:
lgtm
Base Profile Screenshot 2020 03 17 - 20 01 44 84

sorcerer86pt:
also tested what happens when a controlled unit gets released from control after the original player lost

PR #673, opened 2020-03-17
Added a pre-release warning

Mailaender:
like https://github.com/OpenRA/OpenRA/pull/12369.

### DIFF
--- a/mods/ra2/chrome/mainmenu-prerelease-notification.yaml
+++ b/mods/ra2/chrome/mainmenu-prerelease-notification.yaml
@@ -1,0 +1,55 @@
+Background@MAINMENU_PRERELEASE_NOTIFICATION:
+	Logic: PreReleaseWarningPrompt
+	X: (WINDOW_RIGHT - WIDTH) / 2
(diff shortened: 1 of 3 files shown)


pgwipeout:
A bloody impressive work in progress.

The updates in the last few weeks have brought it to the point of being quite playable.

Mailaender:
We will probably be able to ship Red Alert 2 before Tiberian Sun as it has less convoluted engine requirements. Still a lot to do here.

pgwipeout:
True, but there appears to be some overlap (height map comes to mind).

My knowledge of .net is non existent so I can't really assist with code, but I'm happy to keep on testing.

sorcerer86pt:
How to check this warning?

Mailaender:
Type `make version` and start the game.

sorcerer86pt:
well they can be warned now :-) 
Base Profile Screenshot 2020 03 17 - 19 31 10 39

sorcerer86pt:
we should also if possible create a list of know bug/ missing features

Mailaender:
That is what https://github.com/OpenRA/ra2/issues is for.

PR #674, opened 2020-03-17
Fixed anti-air flak weapon discrepancies

Mailaender:
as `InstantHit` does not support inaccuracy.


sorcerer86pt:
also one thinki noticed is that allied anti air missiles are too slow. IFV is not affected.

sorcerer86pt:
lgtm. Flak now always hit, even quick targets.

PR #687, opened 2020-03-22
Make the AI replace harvesters and capture tech buildings

Mailaender:
Adds https://github.com/OpenRA/OpenRA/issues/4561 and https://github.com/OpenRA/OpenRA/issues/15962 here.

### DIFF
--- a/mods/ra2/rules/ai.yaml
+++ b/mods/ra2/rules/ai.yaml
@@ -102,6 +102,7 @@ Player:
 		UnitsToBuild:
 			e1: 90
 			e2: 90
+			engineer: 1
 			dog: 1
 			flakt: 10
 			shk: 10
@@ -123,6 +124,7 @@ Player:
 			tnkd: 15
 			ttnk: 15
 		UnitLimits:
+			engineer: 1
 			dog: 4
 			cmin: 8
 			harv: 8
@@ -137,3 +139,10 @@ Player:
 		RefineryTypes: garefn, narefn
 	BuildingRepairBotModule:
 		RequiresCondition: enable-test-ai
+	CaptureManagerBotModule:
+		RequiresCondition: enable-test-ai
+		CapturingActorTypes: engineer
+		CapturableActorTypes: caoild, caairp, cahosp, cathosp, caoutp
+		CapturableStances: Neutral
+		CheckCaptureTargetsForVisibility: false
+		MaximumCaptureTargetOptions: 15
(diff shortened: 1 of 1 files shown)


MustaphaTR:
There is a reason why CaptureManagerBotModule isn't used in the main mods. The AI is still gonna build that 1 engineer even after it captures everything it can. The trait probably should be revoked to ask the UnitBuilderBotModule to build the engineer, like HarvesterBotModule does, instead of UnitBuilderBotModule building the Engineer by itself and CaptureManagerBotModule finding and using the Idle Engineer.

If you still wanna keep the CaptureManagerBotModule. At least add `engineer` to `ExcludeFromSquadsTypes:` so AI won't attack with them.

pchote:
IMO that really isn't good enough. The fundamental issue is that the AI builds single engineers and then fires the at a tech structure with zero logic as to what is between the barracks and target. A single pillbox/guard tower along the route is enough to turn this feature into a significant AI regression.

This feature needs to be implemented properly, with logic to guard the engineer and path around danger.

Mailaender:
> add `engineer` to `ExcludeFromSquadsTypes:` so AI won't attack with them.

Done.

MustaphaTR:
Maybe a range limit too, so AI won't try to capture something at the other side of the map? But still capture stuff in/really close to its base.

Mailaender:
The AI will target oil derricks in the vicinity first so this works surprisingly well even with such a simple implementation. I doubt we want to copy the `CaptureManagerBotModule` here and try fixing the issue so this is an upstream engine problem.

MustaphaTR:
> I doubt we want to copy the CaptureManagerBotModule here and try fixing the issue so this is an upstream engine problem.

Yes, i just commented here first, because i saw the PR here first.

sorcerer86pt:
ok, the harvester being replaced works (as the AI builds the harvester if one is killed).
My question is that the build order for AI is a bit skewed for refineries. Before it even tries to build vehicle production it always tries to build 4 ore refineries.

It also captured the refinery.

Base Profile Screenshot 2020 03 22 - 13 04 23 59
Base Profile Screenshot 2020 03 22 - 13 10 25 79
Base Profile Screenshot 2020 03 22 - 13 12 43 99
Base Profile Screenshot 2020 03 22 - 13 13 45 40

Mailaender:
Tweaked the rules a bit to combat refinery spam.

sorcerer86pt:
> Tweaked the rules a bit to combat refinery spam.

will test that now.

sorcerer86pt:
The new definitions definitely worked:
Base Profile Screenshot 2020 03 22 - 22 13 53 18
Base Profile Screenshot 2020 03 22 - 22 15 55 15
Base Profile Screenshot 2020 03 22 - 22 15 59 85
Base Profile Screenshot 2020 03 22 - 22 13 48 10


Now after the first refinery either builds a defensive building or a barracks, then after the 2nd refinery it builds the war factory.

Issue #694, opened 2020-03-25
Mirage Tanks can disguise as more than trees


MustaphaTR:
The `MirageTarget:` is defined under `^Tree:` but there are more actors that directly, or indirectly inherit the `^Tree` default that are not actually trees. Signs and Traffic Lights may not look that off but at least `hdstn01` should have the trait removed, imo.

PR #644, opened 2020-02-23
Added the Mirage Tank auto disguise

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/183

PR #707, opened 2020-05-01
Fixed Mirage tank disguising as more than trees

Mailaender:
Followup of https://github.com/OpenRA/ra2/pull/644. Closes https://github.com/OpenRA/ra2/issues/694.


Mailaender:
I think so.

PR #675, opened 2020-03-17
Enabled per-pixel depth calculations

Mailaender:
Incorporate changes from:

* https://github.com/OpenRA/OpenRA/pull/11124 with the help of https://github.com/OpenRA/OpenRA/pull/17720
* https://github.com/OpenRA/OpenRA/pull/11904/
* https://github.com/OpenRA/OpenRA/pull/11091
* https://github.com/OpenRA/OpenRA/pull/11572

PR #708, opened 2020-05-01
Fixed naval sprite offsets

Mailaender:
Followup of #675.


abcdefg30:
This has still some issues: 
grafik
The offset doesn't look correct at all either:
grafik

Mailaender:
Fixed the offsets further.

abcdefg30 (changes requested):
This still doesn't feel quite right compared to the original, where it looks to me like the more or less 3x3 artwork is covering the "upper corner" of the 4x4 footprint like this:
grafik
grafik

Issue #148, opened 2016-02-22
Terror drones


Mailaender:
See http://modenc.renegadeprojects.com/Parasite for reference.

GraionDilach:
Parasite was also used for the Giant Squid and the Firestorm Limpet Drone worked quite similarly as well tbh.

MustaphaTR:
Dogs also uses parasite system but InfantryTypes (targets) are hardcoded to instakilled by a parasite. I literally hate current dog mechanics in OpenRA, in both RA and RA2 mod. It really needs a refactor.

GraionDilach:
`AttackLeap` does need fixes in some terms, since it's - as being the OpenRA equvivalent of `LimboLaunch` - the first step of the parasite system. However, those fixes would be only allowing buildings as valid `AttackLeap` targets and fixing how it ignores `ReloadDelay` atm.

The main difference between RA2 and OpenRA dog logics revolve at speed and that OpenRA bullets don't track the target, unlike the RA2 `Arcing=yes` equivalents, but I guess one can use a missile weapon along with `AttackLeap` to solve that aspect.

Mailaender:
See https://github.com/OpenRA/OpenRA/issues/7506 for the known issues with `AttackLeap`.

ibrokemypie:
+1

Mailaender:
Placeholder added in https://github.com/OpenRA/ra2/pull/423 with infantry only attack.

PR #709, opened 2020-05-01
Added terror drone interaction with vehicles

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/148.


sorcerer86pt:
It's missing one thing: Terror drone infection should be removed if the tank goes to a repair pad, and it starts repairing.

Mailaender:
The drone repair removal will work once the engine is updated to https://github.com/OpenRA/OpenRA/pull/17721.

OpenRA.Mods.RA2/Traits/AttackInfect.cs
```diff
@@ -0,0 +1,105 @@
...
+		[FieldLoader.Require]
+		[Desc("How often to deal the damage.")]
+		public readonly int DamageInterval = 0;
+
+		[Desc("Damage types for the infection damage.")]
+		public readonly BitSet<DamageType> DamageTypes = default(BitSet<DamageType>);
+
+		[Desc("If an external actor delivers more damage than this value, the infector is killed immediately.",
+			"Use -1 to never kill the infector.")]
+		public readonly int SuppressionDamageThreshold = -1;
```
abcdefg30, line 43: What is the usecase for all the `supression` properties here, I don't remember any of that in RA2?

OpenRA.Mods.RA2/Traits/AttackInfect.cs
```diff
@@ -0,0 +1,105 @@
...
+
+			if (self.Location == target.Actor.Location && HasAnyValidWeapons(target))
+				return true;
+
+			return base.CanAttack(self, target);
+		}
+
+		public void GrantJoustCondition(Actor self)
+		{
+			var info = Info as AttackInfectInfo;
```
abcdefg30, line 89: It would likely make sense to expose this as public property, considering we have this cast pattern in multiple places.

Mailaender: Good idea. Done. This is a weird pattern when `Info` is already something else.

Mailaender:
> I still don't quite understand the "joust" terminology, where is the connection to medieval tournaments?

They probably meant diving as in German "hechten", but it isn't in any dictionary? Changed it to _jumping_ now which everyone will understand.

sorcerer86pt:
Plus one for this change, since it improves readability.

A quinta, 23/07/2020, 20:05, Matthias Mailänder <notifications@github.com>
escreveu:

OpenRA.Mods.RA2/Activities/Infect.cs
```diff
@@ -0,0 +1,162 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA2.Activities
+{
+	class Infect : Enter
+	{
+		readonly AttackInfect attackInfect;
+		readonly Target target;
+
+		bool jousting;
```
Phrohdoh, line 26: @Mailaender still `jousting` here

Phrohdoh: Perhaps something like `playingInfectAnim` is more appropriate to this generic (possibly before it needs to be?) trait.

OpenRA.Mods.RA2/Activities/Infect.cs
```diff
@@ -0,0 +1,162 @@
...
+
+			// IsValidInfection validated the lookup, no need to check here.
+			var infectable = targetActor.Trait<Infectable>();
+			return infectable.TryStartInfecting(targetActor, self);
+		}
+
+		protected override bool TryStartEnter(Actor self, Actor targetActor)
+		{
+			var canStartInfect = CanStartInfect(self, targetActor);
+			if (canStartInfect == false)
```
Phrohdoh, line 131:
```suggestion
			if (!canStartInfect)
```

Mailaender:
Updated.

Mailaender:
Rebased.

PR #723, opened 2020-05-24
Renamed superweapon descriptions and added missing notification speeches

Shin-Aska:
I was playing the game on the engine the other day and couldn't help but notice that the superweapon timer names are a bit different compare to the original game. For reference, here is how it looked like in the original.

image

I have also added speech notifications that were missing in them when the superweapon is ready. I'm not sure if the current superweapon descriptions are intentional, so if there is a need for a separate pull request for this other commit, please let me know so I can separate them :)


Mailaender:
Thanks, I suggest https://github.com/OpenRA/ra2/issues/new for your UI findings as it is unrelated to this pull request.

Issue #727, opened 2020-05-25
Incorrect pause animations when player is on low power for certain buildings


Shin-Aska:
I have noticed that some of the buildings in the game right now have incorrect sequences triggering if the player is on lowpower state.
Reference video here: [https://youtu.be/3mGyZ8RIuwo](https://youtu.be/3mGyZ8RIuwo)

Examples: 
Allied construction yard fans stop when its on lowpower (Video: [00:35](https://youtu.be/3mGyZ8RIuwo?t=35)) but on the original game, it doesn't do that (Video: [01:24](https://youtu.be/3mGyZ8RIuwo?t=84)).

Another example is that Air Force Command shouldn't stop its animation even if its on lowpower (Video: [01:56](https://youtu.be/3mGyZ8RIuwo?t=116)) but right now it does (Video: [03:11](https://youtu.be/3mGyZ8RIuwo?t=191))

Finally Ore Purifier should stop its animation when its on lowpower but right now it doesn't (Same video reference timestamps with Air Force Command)

PR #728, opened 2020-05-25
Applied some animation corrections for buildings when player is on low powered.

Shin-Aska:
Modified the following rules:
1. Allied Construction Yard(gacnst)
2. Allied Airforce Command Headquarters (gaairc)
3. Allied Ore Purifier (gaorep)

Changed constructions yard and airforce command to not pause the animation when it is on low power (This is how it works on the original game)
Changed the ore purifier to halt on animation when it is on low power (This is how it works on the original game)

This pull request is related to the issue filed in #727

### DIFF
--- a/mods/ra2/rules/allied-structures.yaml
+++ b/mods/ra2/rules/allied-structures.yaml
@@ -59,7 +59,6 @@ gacnst:
 		ValidPowerStates: Low, Critical
 	WithIdleOverlay@fans:
 		Sequence: idle-fans
-		PauseOnCondition: lowpower
 		RequiresCondition: !build-incomplete
 	WithBuildingPlacedOverlay:
 	ProvidesPrerequisite:
@@ -300,19 +299,15 @@ gaairc:
 		Range: 10c0
 	WithIdleOverlay@top:
 		Sequence: idle-top
-		PauseOnCondition: lowpower
 		RequiresCondition: !build-incomplete
 	WithIdleOverlay@bottom:
 		Sequence: idle-bottom
-		PauseOnCondition: lowpower
 		RequiresCondition: !build-incomplete
 	WithIdleOverlay@mid:
 		Sequence: idle-mid
-		PauseOnCondition: lowpower
 		RequiresCondition: !build-incomplete
 	WithIdleOverlay@lights:
 		Sequence: idle-lights
-		PauseOnCondition: lowpower
 		RequiresCondition: !build-incomplete
 	Targetable:
 		TargetTypes: Ground, C4, DetonateAttack, Structure, SpyInfiltrate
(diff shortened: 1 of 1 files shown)


abcdefg30:
Tbh, not sure if pausing the Airforce HQ isn't a bugfix over the original; will have to think about that. (Doesn't it look a bit weird when that doesn't pause but the Spy Sat does?)

Shin-Aska:
From what I originally understood, if the animation pauses, it was a visual indicator that you can't use it. However, Westwood was inconsistent on this rule because Ore purifier had its animation paused but its passive effect still remains active regardless.

For the airforce command, I think they didn't let it pause because you can still use the aircrafts? But then again probably this is one of the inconsistencies also.

penev92:
1. Thanks for the very detailed research and presentation that you did.
2. At this point saying "sorry for the insane delay" is probably redundant.
3. On the changes:
 - I am definitely with @abcdefg30 on the Airforce HQ disabling being an improvement over the original.
 - I feel a bit less strongly about the ConYard, but it still almost makes sense to me to stop the fans (wouldn't it be cool if we could instead slow them down? Solving the dilemma of "do we stop them if the building still works, although slower due to low power?").
 - As for the Ore Purifier I'm torn because if it continues to work (I wouldn't be surprised if we disable the cash effect too for balance?) it should probably keep the visual effect so the player has *some* feedback...

abcdefg30:
Why slow down instead of stopping?

penev92:
To better reflect what happens to the underlying effect - some structures stop working (radar) and some slow down (production). So we can stop the animation on some and slow down the animation on others to make that clear.

abcdefg30:
Ok that makes sense to me. (As long as it does look good ingame, 👍 )

Shin-Aska:
> So what do people think about slowing down animations on structures whose effects get slowed down on low power to reflect it? (like production structures) This isn't specific to this PR and would affect a bunch more structures and even may not be a bad thing to extend to other mods, btw.

Hi!

Sounds good to me as well, I'll try to apply the changes tonight or tomorrow.

Shin-Aska:
@penev92  and @abcdefg30 i reverted changes #2 and #3 to what it originally was and as for #1, I made some changes base on the discussion where if player is in LOWPOWER state, then it will slow down the fans.

mods/ra2/rules/allied-structures.yaml
```diff
@@ -57,10 +57,12 @@ gacnst:
 	GrantConditionOnPowerState@LOWPOWER:
 		Condition: lowpower
 		ValidPowerStates: Low, Critical
+	WithIdleOverlay@fans@LOWPOWER:
+		Sequence: idle-fans-low-power
+		RequiresCondition: !build-incomplete
```
penev92, line 62: I'm not 100% sure, but this seems like it would play with (under) the other idle animation if not on low power, so I suggest 
```suggestion
		RequiresCondition: !build-incomplete && lowpower
```
to be safe.

penev92 (commented):
I'll need to test, but the way it is set up looks correct.
On a side note, would you be willing to do this for other production structures that have idle (or production) animations?

mods/ra2/rules/allied-structures.yaml
Shin-Aska: Yea, I originally had that condition but removed it thinking it is a bit redundant but I guess it makes more sense to add a little safeguard just in case

Shin-Aska:
> I'll need to test, but the way it is set up looks correct. On a side note, would you be willing to do this for other production structures that have idle (or production) animations?

Sure, I just need to know which structures need this implementation.

penev92:
Actually taking a quick look I can't spot any other candidates.

penev92 (approved):
Looks good to me 👍 
Thank you!

Issue #87, opened 2016-01-23
Only one plane can landing in the Airforce center...


yaaaaa:
In Original RA2 game was 4 planes on 1 Airforce center.

![image](https://cloud.githubusercontent.com/assets/11130958/12531059/e02146ac-c21b-11e5-91f9-78df491a4f6c.png)

scshunt:
This is far from the only issue with harriers, but I think that we would be best served by a proper notion of docking.

pchote:
I suggest adopting the solution from RV ASAP so that @MustaphaTR can claim the bounty and prevent BountySource from taking it for themselves.

> If no Solution is accepted within two years after a Bounty is posted, then the Bounty will be withdrawn and the amount posted for the Bounty will be retained by Bountysource. For Bounties posted before June 30, 2018, the Backer may redeploy their Bounty to a new Issue by contacting support@bountysource.com before July 1, 2020. If the Backer does not redeploy their Bounty by the deadline, the Bounty will be withdrawn and the amount posted for the Bounty will be retained by Bountysource.

Mailaender:
I probably would have already, but I am not sure if a hack with a disabled `Mobile` would be accepted here.

MustaphaTR:
I'm not sure how much that is necessary. I first noticed that dummies were moving around, but wasn't sure why. But i later noticed it was because i removed the passability check from the "Minefield" logic and it was spawning another set of dummies. I fixed that by revering it and making that bits of Airfield passable but didn't remove the don't move hack. I'm not sure if something else can trigger them to move. But tbh the proper fix would be to remove respawning logic and passability checks from "minefield" logic. Then the dummies can probably be a building instead, which can' t move anyway.

Mailaender:
When I experimented with it, the mobile was required so it could spawn onto the existing building.

Mailaender:
Wait, you are right. The mobile hack is indeed not required.

Narann:
Hi all,

We’ve played with the latest RA2 master commit yesterday and I think I can add some more information about plane behavior:

* Planes act like only one place was available (I never saw 2+ planes landed on aircraft).
* When reloading, planes stay on fly and roll for landing one after the other, as if only one place was available and they gently let unloaded planes to land and reload.

On a (secondary) gameplay note, this makes planes really slow to reload.

Keep the good work!

Issue #683, opened 2020-03-21
2 aircraft at the same time land and take off in a neverending loop


Mailaender:
![ra2-aircraft-updown](https://user-images.githubusercontent.com/756669/77229551-111f9800-6b8f-11ea-915b-45e13e218037.gif)

sorcerer86pt:
i saw this many times. It's because the logic for multiple points of landing on the building is missed. It should be calculated for 4, but it only uses the first one.

Phrohdoh:
#735 seems like a mod-level workaround (that I worry we'll end up having to maintain when we don't actually need it) to the engine not supporting multiple landing/takeoff points.

Was it decided that downstream was the proper place to implement this, instead of in the engine?

abcdefg30:
> Was it decided that downstream was the proper place to implement this, instead of in the engine?

Not exactly, #735 is afaik ported from RV which implemented this workaround to not need to wait for the engine's docking code to be rewritten.

Phrohdoh:
> engine's docking code to be rewritten
Seems like the proper fix that we should pursue, if we're going to do anything here.  IME workarounds stay in the codebase even after a proper implementation is available, and if something is a pile of workarounds the maintenance burden grows higher and pool of people willing to work on it diminishes (to potentially even less people than now).

PR #735, opened 2020-06-17
Added more landing pads to the Airforce Command Center

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/87.

### DIFF
--- a/OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
+++ b/OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
@@ -1,0 +1,99 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2020 The OpenRA Developers (see AUTHORS)
+ * This file is part of OpenRA, which is free software. It is made
+ * available to you under the terms of the GNU General Public License
+ * as published by the Free Software Foundation, either version 3 of
+ * the License, or (at your option) any later version. For more
+ * information, see COPYING.
+ */
+#endregion
+
+using System.Collections.Generic;
+using OpenRA.Mods.Common.Traits;
+using OpenRA.Primitives;
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA2.Traits
+{
+	[Desc("This actor places other actors around itself, which keep connected as in they get removed when the parent is sold or destroyed.")]
+	public class SpawnNeighboringActorsInfo : ITraitInfo
+	{
+		[FieldLoader.Require]
+		[ActorReference]
+		[Desc("Types of actors to place. If multiple are defined, a random one will be selected for each actor spawned.")]
+		public readonly HashSet<string> ActorTypes = new HashSet<string>();
+
+		[FieldLoader.Require]
+		[Desc("Locations to spawn the actors relative to the origin (top-left for buildings) of this actor.")]
+		public readonly CVec[] Locations = { };
+
+		public object Create(ActorInitializer init) { return new SpawnNeighboringActors(this, init.Self); }
+	}
+
+	public class SpawnNeighboringActors : INotifyKilled, INotifyOwnerChanged, INotifyActorDisposing, INotifyAddedToWorld, INotifySold
+	{
+		readonly SpawnNeighboringActorsInfo info;
+		readonly List<Actor> actors = new List<Actor>();
+
+		public SpawnNeighboringActors(SpawnNeighboringActorsInfo info, Actor self)
+		{
+			this.info = info;
+		}
+
+		void INotifyAddedToWorld.AddedToWorld(Actor self)
+		{
+			SpawnActors(self);
+		}
+
+		public void SpawnActors(Actor self)
+		{
+			foreach (var offset in info.Locations)
+			{
+				self.World.AddFrameEndTask(w =>
+				{
+					var actorType = info.ActorTypes.Random(self.World.SharedRandom).ToLowerInvariant();
+					var cell = self.Location + offset;
+
+					var actor = w.CreateActor(true, actorType, new TypeDictionary
+					{
+						new OwnerInit(self.Owner),
+						new LocationInit(cell)
+					});
+
+					actors.Add(actor);
+				});
+			}
+		}
+
+		public void RemoveActors()
+		{
+			foreach (var actor in actors)
+				actor.Dispose();
+
+			actors.Clear();
+		}
+
+		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
+		{
+			foreach (var actor in actors)
+				actor.ChangeOwnerSync(newOwner);
+		}
+
+		void INotifyKilled.Killed(Actor self, AttackInfo e)
+		{
+			RemoveActors();
+		}
+
+		void INotifyActorDisposing.Disposing(Actor self)
+		{
+			RemoveActors();
+		}
+
+		void INotifySold.Selling(Actor self) { }
+		void INotifySold.Sold(Actor self)
+		{
+			RemoveActors();
+		}
+	}
+}
--- a/mods/ra2/rules/allied-structures.yaml
+++ b/mods/ra2/rules/allied-structures.yaml
@@ -283,7 +283,7 @@ gaairc:
 		Factions: korea
 	ProvidesPrerequisite@default:
 	Building:
-		Footprint: xxx xxx
+		Footprint: x=x x==
 		Dimensions: 3,2
 	Selectable:
 		Bounds: 142, 150, -3, -42
@@ -320,17 +320,17 @@ gaairc:
 		Amount: -50
 	Reservable:
 	Exit@1:
-		SpawnOffset: -370,370,0
 		Facing: 224
+		SpawnOffset: 1024,512,0
 	Exit@2:
-		SpawnOffset: 370,-370,0
 		Facing: 224
+		SpawnOffset: 512,1024,0
 	Exit@3:
-		SpawnOffset: 1000,370,0
 		Facing: 224
+		SpawnOffset: -512,512,0
 	Exit@4:
-		SpawnOffset: 350,1000,0
 		Facing: 224
+		SpawnOffset: 512,-512,0
 	Production:
 		Produces: Aircraft
 	ProductionBar:
@@ -349,6 +349,20 @@ gaairc:
 	SpawnSurvivors:
 		DeathTypes: ExplosionDeath, BulletDeath
 		Actors: e1, e1
+	SpawnNeighboringActors:
+		ActorTypes: gapad
+		Locations: 1,0, 1,1, 2,1
+
+gapad:
+	Interactable:
+	AlwaysVisible:
+	Reservable:
+	Building:
+		Footprint: x
+		Dimensions: 1,1
+		AllowInvalidPlacement: true
+	Production:
+		Produces: Aircraft
 
 amradr:
 	Inherits: gaairc
--- a/mods/ra2/rules/aircraft.yaml
+++ b/mods/ra2/rules/aircraft.yaml
@@ -194,7 +194,7 @@ orca:
 		Actor: orcahusk
 	HitShape:
 	Rearmable:
-		RearmActors: gaairc, amradr
+		RearmActors: gaairc, amradr, gapad
 
 orcahusk:
 	Inherits: ^PlaneHusk
--- a/mods/ra2/rules/defaults.yaml
+++ b/mods/ra2/rules/defaults.yaml
@@ -976,6 +976,7 @@
 		AltitudeVelocity: 120
 		TakeOffOnCreation: false
 		IdleBehavior: ReturnToBase
+		InitialFacing: 224
 	Hovers@CRUISING:
 		RequiresCondition: cruising
 	WithVoxelBody:


sorcerer86pt:
Ok, tested this commit. 
Some changes need to be done. As you can see in this video:


[multiple airfields](https://drive.google.com/file/d/1wYVb4HsqdW45yZa2V3D_bBagN0hPF5IW/view?usp=sharing)

When multiple airfields exists, the airplanes sometimes enter in loop trying to find a slot, when another one is open (mostly when building new planes, when coming back from bombing mission, they find the slot correctly)

An possible fix is at build time, find a open slot in all airfields that the player controls and then deploy that plane there.

pgwipeout:
Your video is not accessible unfortunately.

sorcerer86pt:
Ok, should be gdrive permissions. I will fix that

A quarta, 24/06/2020, 16:20, Peter Geis <notifications@github.com> escreveu:

sorcerer86pt:
Try now. Fixed the permissions

>

Mailaender:
> An possible fix is at build time, find a open slot in all airfields that the player controls and then deploy that plane there.

Probably not possible with the way this hack works. Only 25% of the pads is constructing aircraft, the others are just for rearming.

yaaaaa:
May be it needs to spawn/build the plane right on free slot, and without flying/jumping after build?

sorcerer86pt:
I approve of that.

A quarta, 24/06/2020, 19:54, yaaaaa <notifications@github.com> escreveu:

MustaphaTR:
Also from the video @sorcerer86pt sent looks like planes land facing 0, they should face 224. https://github.com/MustaphaTR/Romanovs-Vengeance/commit/cc69debf2138312526bcd7c59110544e2a44017b

Mailaender:
Changed the landing facing and added production capabilities to the pads. I don't know why `TakeOffOnCreation: false` is ignored. We have a ticket for that #683.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,86 @@
...
+	[Desc("This actor places other actors around itself.")]
+	public class SpawnNeighboringActorsInfo : ITraitInfo
+	{
+		[FieldLoader.Require]
+		[ActorReference]
+		[Desc("Types of actors to place. If multiple are defined, a random one will be selected.")]
+		public readonly HashSet<string> ActorTypes = new HashSet<string>();
+
+		[FieldLoader.Require]
+		[Desc("Locations to place the mines, from top-left of the building.")]
```
Phrohdoh, line 28:
```suggestion
		[Desc("Locations to spawn the actors, relative to the top-left of the building.")]
```

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,86 @@
...
+	public class SpawnNeighboringActors : INotifyKilled, INotifyOwnerChanged, INotifyActorDisposing
+	{
+		readonly SpawnNeighboringActorsInfo info;
+		readonly List<Actor> actors = new List<Actor>();
+
+		public SpawnNeighboringActors(SpawnNeighboringActorsInfo info, Actor self)
+		{
+			this.info = info;
+
+			SpawnActors(self);
```
Phrohdoh, line 43: Should this be invoked in `OnAddedToWorld` (or something similar), instead of in the trait ctor?
I don't actually know the answer, I'm asking to spark conversation or to learn why that isn't necessary.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		public void RemoveActors()
+		{
+			foreach (var actor in actors)
+				actor.Dispose();
+
+			actors.Clear();
+		}
+
+		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
```
Phrohdoh, line 70: I don't see this documented at all and this seems to be a very specific behavior unrelated to spawning nearby actors.
Regardless of whether this functionality remains, it should be documented IMO.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Phrohdoh: Is `self.Location` _guaranteed_ to be the top-left (and if so, for my edification, how/where)?
If not, I think we should say something like "relative to the origin of this actor."

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,86 @@
...
+			SpawnActors(self);
+		}
+
+		public void SpawnActors(Actor self)
+		{
+			foreach (var offset in info.Locations)
+			{
+				var cell = self.Location + offset;
+				var actor = info.ActorTypes.Random(self.World.SharedRandom).ToLowerInvariant();
+				{
```
Phrohdoh, line 52: Why are we creating another scope here?
Should this be a frame-end task?

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,86 @@
...
+					self.World.CreateActor(true, actor, new TypeDictionary
+					{
+						new OwnerInit(self.Owner),
+						new LocationInit(cell)
+					});
+				}
+			}
+		}
+
+		public void RemoveActors()
```
Phrohdoh, line 62: Same here in general regarding undocumented behavior.

mods/ra2/rules/allied-structures.yaml
```diff
@@ -349,6 +349,20 @@ gaairc:
 	SpawnSurvivors:
 		DeathTypes: ExplosionDeath, BulletDeath
 		Actors: e1, e1
+	SpawnNeighboringActors:
+		ActorTypes: gapad
+		Locations: 1,0, 1,1, 2,1
+
+gapad:
```
Phrohdoh, line 356: Why, technically, are we splitting this into multiple actor types?
Is this required to work around an engine issue?
If not, to me this seems like (prematurely) building more than we need.

Phrohdoh:
`SpawnNeighboringActors` from a modding perspective (given the information this PR as-is makes available to modders) _seems_ awfully similar to `FreeActor`, so I believe differentiating via docs/name would be beneficial.

mods/ra2/rules/allied-structures.yaml
Mailaender: There can't be more than one docking activity per actor. This is an engine limitation which can't be resolved here.

Mailaender:
Updated and fixed an issue where selling the aircraft command would not remove the additional pads.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: I think you are right and changed the code accordingly. This probably fixes nothing other than unnoticeable timing.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: Added a note to the trait description.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: Yes, that was actually broken probably from an incomplete copy and paste. Fixed.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: Look at the `Building` code or place an actor in the editor. The origin is always top-left. Removed the legacy cruft. This was used to spawn mines as in Generals China ability before and recycled.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,97 @@
...
+using OpenRA.Traits;
+
+namespace OpenRA.Mods.RA2.Traits
+{
+	[Desc("This actor places other actors around itself, which keep connected.")]
+	public class SpawnNeighboringActorsInfo : ITraitInfo
+	{
+		[FieldLoader.Require]
+		[ActorReference]
+		[Desc("Types of actors to place. If multiple are defined, a random one will be selected.")]
```
phrohdoh, line 24:
```suggestion
		[Desc("Types of actors to place. If multiple are defined, a random one will be selected for each actor spawned.")]
```

To make it more clear that each spawned actor will be a random type from this set, instead of a random single type being selected for all to-be-spawned actors.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
phrohdoh: I've switched this comment back to unresolved since this still is not documented as far as I can tell.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
phrohdoh: Is that in reference to the "which keep connected?"
That makes me think this'd be used for (visually) connecting walls, not ownership of the spawned actors always being that of the actor that spawned them throughout the game.

Reviewing this PR made me realize that the trait it introduces into RA2 seems like `FreeActor` with cell position offsets (instead of world) and selling, killing, and ownership changing of the free actors.
Would it make any sense to use (upgrading the implementation as necessary) multiple `FreeActor`s instead of a single similar `SpawnNeighboringActors`?

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,97 @@
...
+		{
+			foreach (var actor in actors)
+				actor.Dispose();
+
+			actors.Clear();
+		}
+
+		void INotifyOwnerChanged.OnOwnerChanged(Actor self, Player oldOwner, Player newOwner)
+		{
+			foreach (var actor in actors)
```
phrohdoh, line 77: Do we need to check for exists/alive/in-world here (or does owner change machinery handle that for us?) and/or keep this list up-to-date (removing from the list when referenced actors die or enter/exit world)?

This trait is clearly intended to be used for spawning buildings (the aircraft pads in specific), so would it make sense to bake that intention into the code (no need to be overly generic IMO, especially if we aren't going to consider all of the possible interactions and events we need to think about)?

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,97 @@
...
+				actor.ChangeOwnerSync(newOwner);
+		}
+
+		void INotifyKilled.Killed(Actor self, AttackInfo e)
```
phrohdoh, line 81: Undocumented behavior.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,97 @@
...
+			RemoveActors();
+		}
+
+		void INotifyActorDisposing.Disposing(Actor self)
+		{
+			RemoveActors();
+		}
+
+		void INotifySold.Selling(Actor self) { }
+		void INotifySold.Sold(Actor self)
```
phrohdoh, line 92: Undocumented behavior.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,97 @@
...
+		[Desc("Locations to spawn the actors relative to the origin (top-left for buildings) of this actor.")]
+		public readonly CVec[] Locations = { };
+
+		public object Create(ActorInitializer init) { return new SpawnNeighboringActors(this, init.Self); }
+	}
+
+	public class SpawnNeighboringActors : INotifyKilled, INotifyOwnerChanged, INotifyActorDisposing, INotifyAddedToWorld, INotifySold
+	{
+		readonly SpawnNeighboringActorsInfo info;
+		readonly List<Actor> actors = new List<Actor>();
```
phrohdoh, line 37: I don't see items being added to this list (which could be a hashset? not sure if that'll really matter).
Given that, I don't understand how this trait actually keeps ownership/existence in sync.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
```diff
@@ -0,0 +1,97 @@
+#region Copyright & License Information
+/*
+ * Copyright 2007-2020 The OpenRA Developers (see AUTHORS)
```
phrohdoh, line 3: Not entirely related to this PR in specific, but should copyright of non-engine/core code be attributed to engine/core?  Also copy+pasting `2007-<year>` around doesn't actually make that the case.

mods/ra2/rules/allied-structures.yaml
```diff
@@ -349,6 +349,20 @@ gaairc:
 	SpawnSurvivors:
 		DeathTypes: ExplosionDeath, BulletDeath
 		Actors: e1, e1
+	SpawnNeighboringActors:
+		ActorTypes: gapad
+		Locations: 1,0, 1,1, 2,1
```
phrohdoh, line 354: Why only 3?
This seems to line up, in some way I don't yet understand, with the footprint having 3 `=`s, not 4 (or even 6 `x`s).

mods/ra2/rules/allied-structures.yaml
```diff
@@ -349,6 +349,20 @@ gaairc:
...
+
+gapad:
+	Interactable:
+	AlwaysVisible:
+	Reservable:
+	Building:
+		Footprint: x
+		Dimensions: 1,1
+		AllowInvalidPlacement: true
```
phrohdoh, line 363: Why?

mods/ra2/rules/allied-structures.yaml
```diff
@@ -283,7 +283,7 @@ gaairc:
 		Factions: korea
 	ProvidesPrerequisite@default:
 	Building:
-		Footprint: xxx xxx
+		Footprint: x=x x==
```
phrohdoh, line 286: Why this footprint in particular?

phrohdoh (changes requested):
Inline comments regarding undocumented behavior, making the trait only as abstract/generic as necessary (this isn't engine/core where we have and expect 3rd-party reuse (and if that happens, things can be made more generic then)), how the trait logic works in general (re not seeing items added to the `actors` instance member), and some MiniYaml values.

mods/ra2/rules/allied-structures.yaml
Mailaender: The main building is already one pad.

mods/ra2/rules/allied-structures.yaml
Mailaender: So it can be placed on top of the main building.

mods/ra2/rules/allied-structures.yaml
Mailaender: The first `x´ is the pad that comes with the main building. The 3 `=` are needed so another actor can be placed on top of it providing 1 more pad each.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: I won't change just this file for consistency.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: Updated.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: That is a bug. Fixed below.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: I explained "they keep connected" more explicitly now.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: I think not as this trait is really used for these invisible invincible fake actors.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
Mailaender: Well kinda.

Mailaender:
Rebased and updated.

OpenRA.Mods.RA2/Traits/SpawnNeighboringActors.cs
phrohdoh: Owner changing is still undocumented.

MustaphaTR:
Needs rebase.

abcdefg30:
Tbh I'm not sure if this PR is necessary. If this repository seems unmaintained to people, I don't think there is a reason to add this workaround, or is there? I.e. since we're not in a rush we can wait for docking to work like we need.

Issue #710, opened 2020-05-02
Terrain bleeding through the large building sprites


Mailaender:
![Imgur](https://i.imgur.com/7j91jI8.png)

pchote:
<img width="625" alt="Screenshot 2020-05-02 at 12 48 32" src="https://user-images.githubusercontent.com/167819/80863316-4c73b300-8c73-11ea-8e52-f244919defee.png">

Something (possibly an engine-level bug) is causing the soviet conyard to use the giant squid sprite sheet as its depth data.

pchote:
The depth sprite handling was rewritten between `release-20200202` and the current playtest series / bleed, so it is possible that this has already been fixed.

Mailaender:
It still happens on https://github.com/OpenRA/ra2/pull/738 but the soviet war factory is now the worse offender. Changing `DepthSpriteOffset` makes other sprite patterns appear. Picking a one that mimics the sprite offset fixes/hides the problem.

pchote:
I put together a testcase based on #778, https://github.com/OpenRA/OpenRA/pull/19561, and the fix https://github.com/OpenRA/OpenRA/pull/19563.

Check out https://github.com/OpenRA/ra2/commit/582911b48b04db20359c34c8655e29f9bbd42999 and start a skirmish as Cuba on the top-right spawn of "The Alamo" map (or don't, but this should guarantee that you trigger the bug). Disable the shroud and make sure you are zoomed relatively close (zooming too far out changes the order that sheets are mapped and may hide the bug) before enabling the depth preview and tweaking the contrast/offset until you are satisfied that you see the problem:

<img width="945" alt="Screenshot 2021-07-25 at 11 46 49" src="https://user-images.githubusercontent.com/167819/126897339-60df191f-206a-468c-b277-a6918dc200f0.png">

Its not as obvious as seeing a giant squid embedded in the depth map, but is still clearly not right.

Now repeat with https://github.com/OpenRA/ra2/commit/8bc94038d58c227deccab0fdea2eb7755f7f2f26, which adds https://github.com/OpenRA/OpenRA/pull/19563 and observe that the conyard now displays correctly:

<img width="1149" alt="Screenshot 2021-07-25 at 11 57 28" src="https://user-images.githubusercontent.com/167819/126897366-dc9b7fbb-842e-438b-bdd3-035ecdd7f457.png">

The terrain is still wrong, but that's an unrelated RA2 yaml problem.

[...]

PR #747, opened 2020-09-13
Introduce Depth Sprite Offsets

Mailaender:
See https://github.com/OpenRA/OpenRA/pull/7968


phrohdoh:
This doesn't appear to _fix_ #710 to me.
Screen Shot 2021-01-02 at 11 18 57 AM
Screen Shot 2021-01-02 at 11 18 44 AM

Mailaender:
That was probably too optimistic. Linked the more appropriate pull request that this mirrors.

pchote:
The current default offsets are already mostly correct. Adding the large negative Y offset makes most of the buildings run off the top edge of the depth sprite and use whatever arbitrary texture data is there - actually introducing more cases of #710!

Before (notice that the battle lab actually overruns already, ra2 will need to create a taller isodepth file to fix this!):

Screenshot 2021-07-25 at 14 16 31

After:
Screenshot 2021-07-25 at 14 15 17

PR #750, opened 2020-10-10
Added support for Visual Studio code.

Mailaender:
This is the counterpart to https://github.com/OpenRA/OpenRA/pull/18280.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,19 @@
+{
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"name": "ra2",
+			"type": "clr",
+			"linux": {
+				"type": "mono"
+			},
+			"osx": {
+				"type": "mono"
+			},
+			"request": "launch",
+			"program": "${workspaceRoot}/engine/OpenRA.Game.exe",
+			"cwd": "${workspaceRoot}",
+			"args": ["Engine.LaunchPath=${workspaceRoot}", "Engine.ModSearchPaths=${workspaceRoot}/mods,./mods", "Game.Mod=ra2", "Debug.DisplayDeveloperSettings=true"]
+		},
+	]
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,7 @@
+{
+	"recommendations": [
+		"ms-dotnettools.csharp",
+		"EditorConfig.EditorConfig",
+		"ms-vscode.mono-debug"
+	]
+}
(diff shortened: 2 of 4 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
...
+            "linux": {
+                "type": "mono"
+            },
+            "osx": {
+                "type": "mono"
+            },
+            "request": "launch",
+            "program": "${workspaceRoot}/engine/OpenRA.Game.exe",
+            "cwd": "${workspaceRoot}",
+            "args": ["Engine.LaunchPath=${workspaceRoot}", "Engine.ModSearchPaths=${workspaceRoot}/mods", "Game.Mod=ra2", "Debug.DisplayDeveloperSettings=true"]
```
abcdefg30, line 16: Shouldn't the `ModSearchPaths` include just `./mods` here as well? (See https://github.com/OpenRA/ra2/blob/master/launch-game.sh#L21)

abcdefg30 (commented):
`launch.json` and `tasks.json` should be converted to use tabs instead of spaces to be consistent with the other files. Lgtm otherwise.

.vscode/launch.json
Mailaender: `${workspaceRoot}/mods` probably translates to just `./mods` already.

Mailaender:
Nope, these are (Visual Studio Code) `.json` files. Just like `.travis.yml` it should be indented in upstream way.

abcdefg30:
Half of the files here are already indented using tabs, and OpenRA/OpenRA#18280 has all files indented with tabs as well.

.vscode/launch.json
abcdefg30: `./mods` will be in the engine directory, won't it?

Mailaender:
Okay, if tabs don't count as syntax errors then here you go.

.vscode/launch.json
Mailaender: I think that is relative to the repository root so it will be the `mods` folder where `ra2` is in.

abcdefg30: Then we'd be missing the mods folder of the engine directory or is that included in some other way?

Mailaender: I can only say that is the exact same setup I use for @OpenHV and I never encountered problems during debugging.

abcdefg30: I'm just wondering since we include that in the start scripts (see link in the first comment). Perhaps the `modcontent` mod?

Mailaender: Maybe. I never debugged during content installation.

abcdefg30: What is the status here. Does this work with the content mod or not?

Mailaender: Amended it so it does work now.

abcdefg30 (approved):
Thanks!

Issue #719, opened 2020-05-10
Wrong exit offsets on Aircraft Carrier and Destroyer


abcdefg30:
![destOffset](https://user-images.githubusercontent.com/7704140/81499453-16b27800-92cc-11ea-89ee-4cb0b741ac9c.gif)
See https://github.com/OpenRA/ra2/pull/703#issuecomment-626335284.
This will likely have to be fixed on the `Exit` trait upstream.

PR #752, opened 2020-11-16
Correct offset for ship spawned aircraft #719

maxrwalker:
Rotated the offset vector of the child (plane) by the facing of the parent (ship)

### DIFF
--- a/OpenRA.Mods.RA2/Traits/BaseSpawnerParent.cs
+++ b/OpenRA.Mods.RA2/Traits/BaseSpawnerParent.cs
@@ -202,14 +202,17 @@ namespace OpenRA.Mods.RA2.Traits
 		public virtual void SpawnIntoWorld(Actor self, Actor child, WPos centerPosition)
 		{
 			var exit = self.RandomExitOrDefault(self.World, null);
-			SetSpawnedFacing(child, self, exit);
+			SetSpawnedFacing(child, exit);
 
 			self.World.AddFrameEndTask(w =>
 			{
 				if (self.IsDead)
 					return;
 
-				var spawnOffset = exit == null ? WVec.Zero : exit.Info.SpawnOffset;
+				var spawnOffset = WVec.Zero;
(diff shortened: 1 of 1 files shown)


maxrwalker:
BTW - I saw the comment about the Exit trait needing modification - but I think this works?

OpenRA.Mods.RA2/Traits/BaseSpawnerParent.cs
```diff
@@ -202,14 +202,17 @@ void INotifyActorDisposing.Disposing(Actor self)
...
 
 			self.World.AddFrameEndTask(w =>
 			{
 				if (self.IsDead)
 					return;
 
-				var spawnOffset = exit == null ? WVec.Zero : exit.Info.SpawnOffset;
+				var spawnOffset = WVec.Zero;
+				if (exit != null)
+					spawnOffset = exit.Info.SpawnOffset.Rotate(new WRot(WAngle.Zero, WAngle.Zero, WAngle.FromFacing(64 - facing.Facing)));
```
maxrwalker, line 214: I could move this to the Exit trait as that appears to maintain the facing?

abcdefg30: I think that would indeed be better. Where is the magic `64 -` coming from?

phrohdoh:
@maxrwalker would you please address [abcdefg30's comment](https://github.com/OpenRA/ra2/pull/752#discussion_r527671493) so that we can move forward with this PR?

OpenRA.Mods.RA2/Traits/BaseSpawnerParent.cs
Mailaender: I can confirm that this also works in a 2D plane with only sprites. Just had to change offset from `-250` to `250` so the minus sign may be "wrong" compared to `Armament` `LocalOffset` from a regular bullet.

PR #776, opened 2021-06-19
Fix the Tesla Tank's speed

### LINKED ISSUE #775: Tesla Tanks are slower than they should
Wanted to note this here so i won't forget. In the original ini (i checked both RA2 and YR to be sure it is not a difference between 2) it seems like they have `Speed=6`, which is same as Rhino and should be 90 here. But seems like this mod uses 75.

abcdefg30:
Closes #775.
Confirmed that `rules.ini` lists the same speed for Rhinos and Tesla Tanks.


MustaphaTR (approved):
Looks good but can't merge. Something is wrong with the checks.

Issue #762, opened 2021-01-10
Allow Spy to Disguise as Certain Units Without Crash


WilliamWsyHK:
I tried to disguised the Spy as Rocketeers, but of course it would fail because IIRC in the original game it just pretend to be walk on the ground, but it gives exception instead.
```
Exception of type `System.InvalidOperationException`: Unit `spy` does not have a sequence named `fly`
    at OpenRA.Graphics.SequenceProvider.GetSequence(String unitName, String sequenceName)
    at OpenRA.Graphics.Animation.PlaySequence(String sequenceName)
    at OpenRA.Graphics.Animation.PlayRepeating(String sequenceName)
    at OpenRA.Mods.Common.Traits.Render.WithInfantryBody.Tick(Actor self)
    at OpenRA.WorldUtils.DoTimed[T](IEnumerable`1 e, Action`1 a, String text)
    at OpenRA.World.Tick()
    at OpenRA.Game.InnerLogicTick(OrderManager orderManager)
    at OpenRA.Game.LogicTick()
    at OpenRA.Game.Loop()
    at OpenRA.Game.Run()
    at OpenRA.Game.InitializeAndRun(String[] args)
    at OpenRA.Program.Main(String[] args)
```

We should be allowed for spy to disguise as certain units, without any crashes.
(Disguise ready)

vgaming:
Confirming, I just got a similar
```
Exception of type `System.InvalidOperationException`: Unit `spy` does not have a sequence named `idle-air`
  at OpenRA.Graphics.SequenceProvider.GetSequence (System.String unitName, System.String sequenceName) [0x00057] in <2f36d06c9a324d228639d627cc054bbb>:0 
  at OpenRA.Graphics.Animation.GetSequence (System.String sequenceName) [0x0000c] in <2f36d06c9a324d228639d627cc054bbb>:0 
  at OpenRA.Graphics.Animation.PlaySequence (System.String sequenceName) [0x00000] in <2f36d06c9a324d228639d627cc054bbb>:0 
```

Mailaender:
https://github.com/OpenRA/OpenRA/pull/17742 might have already fixed that.

mfzl:
Any workarounds for this, we just had a crash middle of a game.

On a tangential note, how does one get started with development? I didn't find too much information in the book and contribution guide.

Mailaender:
There is a patch at https://github.com/OpenRA/ra2/pull/786 which has not yet been integrated.

[...]

PR #786, opened 2021-10-21
Fixed spies disguising as rocketeers crashing the game

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/762


Mailaender:
It does not make any sense to me because you could easily spot a grounded rocketeer as a spy, but okay.

Issue #406, opened 2018-01-20
US Spy Sat doesn't seem to work.


beigexperience:
You might know this already but the spy sat radar map revealer of the Americans doesn't seem to work. 

I have absolutely no idea what I'm doing but changing the basic `RevealsShroud` Range to a big number seems to mimick the desired effect until it bumps up against `MaximumTileSearchRange` from `OpenRA.Game/Map/MapGrid.cs`. This is probably not the proper solution.

I've just started reading through the code, but as far as I understand you'd want to encapsulate the shroud revealing logic in some sort of "trait" I guess you call it that's invoked on power gained to the building/power loss right?

abcdefg30:
We probably want to abuse the `GPSPower` or create a similar trait, yes.

GraionDilach:
Do note that fog of war is the reason why RA1 got GPS icons instead of the full-map-reveal it had. I'm not sure if recreating the shroud/FoW removal entirely wiith spysats would be good.

beigexperience:
Hmm, the visual studio .sln solution file seems to be broken?

Nihilianth:
It seems to me like this one can only be fixed with changes to Shroud.cs in  OpenRA (e.g. splitting the disable flag to work for shroud and fog separately or adding a new SourceType). It seems impossible atm to temporarily disable the shroud without also disabling fog of war and affecting exploration data permanently.

sorcerer86pt:
Well in RA2 OG, GPS satellite it would reveal all, and it was a major advantage in both single and multiplayer game. I don't know if OpenRA decided to GPS icons for multiplayer balance, or if it was like that in the original RA game. 

RA2 GPS had it's shortcomings and advantages of course in relation to RA1. Since it was based on a building, it was deployed instantly if we had the energy for it, but if it was not powered or destroyed that effect was disabled. 
In RA1 we had to wait to finish deploying the satellite, but if the tech center was destroyed, the effect was still in usage, as long we had the radar some. 

https://cnc.fandom.com/wiki/Global_Positioning_System

IMHO,  aside from a balance issue, we should mantain the OG behavior.

In relation to implementation, the spy satellite building should have a trait, that as long the building was powered, it revealed all fog and map. This would implement the OG behavior, mantain some balance and encapsulated the behavior for other mods.

GraionDilach:
[...]

PR #802, opened 2022-12-31
Added Spy Satellite map reveal

Mailaender:
Closes https://github.com/OpenRA/ra2/issues/406


MustaphaTR (approved):
In the original game losing SpySat reshrouded the map which doesn't seem to happen here. Of course we have FoW here by default, but still would be an issue if the player has it disabled. I think that's out of scope of this PR tho, so approving, but perhaps the issue should be kept open or a new one should be opened on that.

PR #811, opened 2025-04-11
Update mod to release-20231010

penev92:
A slightly belated update.

The things I had issues with:
- The SpriteSequence update - I had to modify the `ExplicitSequenceFilenames` to handle RA2's specific naming - _I have no guarantees that went with no issues_
- The `tib02` and `tib14` sequences were crashing after the update because they manually defined frame indices
- I haven't updated the TFD package installer because that might need to be split into two ContentPackages and I can't test it


penev92:
OK, I finally got the CI to pass. I'm hoping that we can go through this PR relatively quickly so I can do an update to `release-20250330` in a timely manner.

Mailaender (approved):

PR #814, opened 2025-04-23
Update mod to release-20250330

penev92:
Two things to note:
- The `ChronoResourceDelivery` here is broken because of the ongoing docking rework upstream , but I don't think it's worth adapting to the current WIP engine APIs until those settle to a semi-stable state.
- The `ImportRA2MapCommand` may have some regressions (I believe it doesn't import lamps after this update), but it deserves a full rewrite (as mentioned in https://github.com/OpenRA/ra2/pull/798#issuecomment-1495593937) based on https://github.com/OpenRA/OpenRA/pull/20784, which I am already doing in a follow-up branch and would like to keep in a separate PR because this one is big enough.

PR #815, opened 2025-05-06
Rewrote ImportRA2MapCommand to use the upstream ImportGen2MapCommand

penev92:
This has been a very long time coming, but it is finally here. The follow-up to https://github.com/OpenRA/OpenRA/pull/20784.
Might be good to reimport the current maps, but not strictly necessary.

Depends on #814.


OpenRA.Mods.RA2/UtilityCommands/ImportRA2MapCommand.cs
```diff
@@ -337,437 +338,25 @@ sealed class ImportRA2MapCommand : IUtilityCommand
...
-		};
-
-		static readonly Dictionary<string, string> ReplaceActors = new()
+		// Dogs obviously don't deploy into different dogs, but there used to be a "ReplaceActors"
+		// because in OG RA2 there were two separate dog actors and we just use faction-specific sprites.
+		// Since what this does is replace actor A with actor B
+		// and add a DeployStateInit and that init doesn't hurt,
+		// we piggy-back off of it to translate to the proper actor time,
+		// even if with a redundant DeployStateInit.
+		protected override Dictionary<string, string> DeployableActors { get; } = new()
```
Mailaender, line 347: Then simply keep this forked version instead inheriting from the TS artillery function. Deploying dogs is awkward.

